PR #294, opened 2012-10-21
Add a step to cowboy_rest to support HTTP 429.

betawaffle:
See [rfc6585](http://tools.ietf.org/html/rfc6585#section-4) for an explaination of status code 429.

The callback for that example might look something like this:

``` erlang
-define(RATE_LIMIT_MSG,
        "I only allow 50 requests per hour "
        "to this Web site per logged in user.  "
        "Try again soon.").

rate_limited(Req, State) ->
    ReqCount = user_requests(Req, 3600),
    if
        ReqCount > 50 ->
            Body = error_html(<<"Too Many Requests">>,
                              <<?RATE_LIMIT_MSG>>),
            {ok, Req2} = cowboy:set_resp_body(Body, Req),
            {{true, 3600}, Req2, State};
        true ->
            {false, Req, State}
    end.

error_html(Title, Message) ->
    <<"<html>"
      "<head>"
      "<title>", Title/binary, "</title>"
      "</head>"
      "<body>"
      "<h1>", Title/binary, "</h1>"
      "<p>", Message/binary, "</p>"
      "</body>"
      "</html>">>.
```

Bring on the feedback!

### DIFF
--- a/src/cowboy_rest.erl
+++ b/src/cowboy_rest.erl
@@ -82,7 +82,24 @@ upgrade(_ListenerPid, Handler, Opts, Req) ->
 	end.
 
 service_available(Req, State) ->
-	expect(Req, State, service_available, true, fun known_methods/2, 503).
+	expect(Req, State, service_available, true, fun rate_limited/2, 503).
+
+%% rate_limited/2 should return a false, true, or {true, RetryAfter}
+%% with RetryAfter being the number of seconds the client should wait.
+rate_limited(Req, State) ->
+	case call(Req, State, rate_limited) of
+		no_call ->
+			known_methods(Req, State);
+		{halt, Req2, HandlerState} ->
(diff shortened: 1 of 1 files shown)


tnt-dev:
Any news on this?

betawaffle:
@tnt-dev I have no immediate plans to work on this. It would have to come from someone else.

essen:
This is something that I want to add in Cowboy 2 or soon after that, depending on time constraints.

essen:
I've implemented this in my local copy. Compared to this it has been moved after is_authorized/forbidden so that we know the user can access the resource to begin with, and I've removed the empty true clause as I think sending retry-after is important. I also have added tests and updated the docs.

I'll push it later today and it'll be part of 2.6. Closing, thanks!

PR #514, opened 2013-04-24
Create cowboy_static_sendfile_opts.erl to support file:sendfile/5 chunk_size option

sgdreamer7:
Modification of cowboy_static.erl to reduce memory consumption using option chunk_size of file:sendfile/5. By default, cowboy_static use file:sendfile/2 with chunk_size of 20 MB per file. This cause in my project memory consumption grow by 200-300 MB during download an svg files with included ecmascript files.


essen:
Closing, see linked ticket for followup. This will be planned for after 2.0. Thanks!

PR #596, opened 2013-09-27
Added charset option to cowboy_static

koctep:


essen:
5 years later, I have added a `charset` option to cowboy_static. It's modeled like the `mimetypes` option and a function can be used to detect a file's charset. Will push soon. Thanks!

koctep:
thank you

PR #681, opened 2014-03-12
Add cowboy_req:buffer_data/3

ferd:
This function listens for and inspects its buffer until it has received
some amount of data (if any) and then returns.

This can be useful when, for example, multiple connections might be
listened to at once with a timeout to serve the first one to answer, or
in similar cases where work can be done while waiting for data on the
line.

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -77,6 +77,7 @@
 -export([init_stream/4]).
 -export([stream_body/1]).
 -export([stream_body/2]).
+-export([buffer_data/3]).
 -export([skip_body/1]).
 -export([body/1]).
 -export([body/2]).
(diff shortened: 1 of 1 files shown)


essen:
I don't think I will ever merge this. The way the current body function works, you already can get data in a similar way. You just have to keep it around in your code.

The proxy use case in Cowboy 2 will be covered by the stream interface, not by traditional handlers (though your custom code may still use handlers), and the stream interface will automatically forward and/or buffer data depending on configuration (even if you use only normal handlers), so it doesn't look like it will be needed there either.

Closing, thanks!

ferd:
That isn't necessarily true in the case where pipelined requests are waiting down the line, but you still want to see if the connection is open.

The way of doing that is to get all the content out of the TCP receive buffer to let the stack report FIN packets once the content has been consumed. Just waiting on the body means the pipelined requests may artificially hold the connection open longer than it needs to be.

But I mean it's fine, we ended up going with our `cowboyku` fork for all these fixes anyway. Our proxy requirements just don't match a server's requirement plainly enough in many places.

essen:
I'm guessing you didn't see my talk (or slides) about Cowboy 2. :-)

The only way in Cowboy 2 for a request process to prevent HTTP/1.1 pipelined requests from running is if its request body is larger than what flow control currently allows. So should be just about the same as what you got.

ferd:
Yeah, we had needed it in 2014 though :)

As I fine, this is totally fine to have closed.

PR #752, opened 2014-10-08
change init/2 to return #state{} in documentation

sstrigler:
Most examples returned 'Opts' as given by second argument to init. By
using '#state{}' the examples make it more clear that this is what is
being passed as 'State' to all subsequent callbacks (if any).


essen:
Merged, thanks!

PR #766, opened 2014-11-19
Add an index.html that is also served statically as the default for /

davidw:


examples/static_world/src/static_world_app.erl
```diff
@@ -13,8 +13,9 @@
 start(_Type, _Args) ->
 	Dispatch = cowboy_router:compile([
 		{'_', [
+			{"/", cowboy_static, {priv_file, static_world, "index.html"}},
 			{"/[...]", cowboy_static, {priv_dir, static_world, "",
-				[{mimetypes, cow_mimetypes, all}]}}
+							[{mimetypes, cow_mimetypes, all}]}}
```
essen: Only seeing whitespace change on this line, please remove the change.

essen:
Good but see point mentioned above, plus make that a single commit. Thanks!

essen:
Merged, thanks!

PR #769, opened 2014-12-04
Add typespecs for 'state' record in cowboy_spdy module

sile:
Added some typespecs to support `warn_untyped_record` compiler option.

```bash
# warning example
$ erlc +warn_untyped_record src/cowboy_spdy.erl 
src/cowboy_spdy.erl:55: Warning: record state has field(s) without type information
```


essen:
Merged and added the warning to the list of options Cowboy compiles with. Thanks!

PR #776, opened 2014-12-19
A bit more informative errors of constraints

manifest:
Currently, we can't determine which of constraint fails and what a field for. in case the constraint returns `false`, we receive a not informative `{reason, case_clause}`. And when the constraint specified without a default value and a field isn't presented in the request, we receive `{reason,bad_key}`. We could make this behaviour a bit more informative by passing a field name and a constraint function with the exception `{bad_constraint, Key | {Key, Constraint}}`.
- `{bad_constraint, {Key, Constraint}}`  
  when the constraint returns 'false' and a field is presented in the request
- `{bad_constraint, Key}`  
  when the constraint specified without a default value and a field isn't presented in the request

``` erlang
%% Example

byte_size_eq(Size) ->
  fun         
    (Val) when byte_size(Val) =:= Size -> true;
    (_)                                -> false
  end.  

malformed_request(Req, State) ->
  ByteSizeEq1 = byte_size_eq(1),
  Constraints =
    [ {a, int, 42},
      {b, ByteSizeEq1} ],

  ErrorDesc =
    fun
      ({Key, F}) ->
        Field = atom_to_binary(Key, utf8),
        case F of
          int         -> <<"The '", Field/binary, "' field must be an integer.">>;
          ByteSizeEq1 -> <<"The '", Field/binary, "' field must be a bynary string of length 1.">>;
          _           -> <<"Invalid value of the '", Field/binary, "' field.">>
        end;
      (Key) ->
        <<"The '", (atom_to_binary(Key, utf8))/binary, "' field is required.">>
    end, 

  try cowboy_req:match_qs(Constraints, Req) of
    M ->
      {false, Req, M}
  catch
    _:{bad_constraint, Reason} ->
      error_logger:error_report([bad_constraint, {reason, Reason}]),
      Req2 = cowboy_req:set_resp_body(jsxn:encode(#{error => ErrorDesc(Reason)}), Req),
      Req3 = cowboy_req:set_resp_header(<<"content-type">>, <<"application/json">>, Req2),
      {true, Req3, State}
  end.
```

``` sh
## Case 1

curl -4i -XGET "http://localhost:8081/?a=A"
HTTP/1.1 400 Bad Request
connection: keep-alive
server: Cowboy
date: Fri, 19 Dec 2014 13:59:39 GMT
content-length: 45
content-type: application/json

{"error":"The 'a' field must be an integer."}
```

``` erlang
=ERROR REPORT==== 19-Dec-2014::16:59:39 ===
    bad_constraint
    reason: {a,int}
```

``` sh
## Case 2

curl -4i -XGET "http://localhost:8081/?a=1"
HTTP/1.1 400 Bad Request
connection: keep-alive
server: Cowboy
date: Fri, 19 Dec 2014 14:00:17 GMT
content-length: 32
content-type: application/json

{"error":"The 'b' field is required."}
```

``` erlang
=ERROR REPORT==== 19-Dec-2014::17:00:18 ===
    bad_constraint
    reason: b
```

``` sh
## Case 3

curl -4i -XGET "http://localhost:8081/?a=1&b=BB"
HTTP/1.1 400 Bad Request
connection: keep-alive
server: Cowboy
date: Fri, 19 Dec 2014 14:00:50 GMT
content-length: 62
content-type: application/json

{"error":"The 'b' field must be a bynary string of length 1."
```

``` erlang
=ERROR REPORT==== 19-Dec-2014::17:00:50 ===
    bad_constraint
    reason: {b,#Fun<example_hello_handler.0.73648182>}
```

``` sh
## Case 4

curl -4i -XGET "http://localhost:8081/?a=1&b=B"
HTTP/1.1 200 OK
connection: keep-alive
server: Cowboy
date: Fri, 19 Dec 2014 14:01:16 GMT
content-length: 15
content-type: application/json

{"a":1,"b":"B"}
```

### DIFF
--- a/src/cowboy_constraints.erl
+++ b/src/cowboy_constraints.erl
@@ -16,10 +16,12 @@
 
 -export([validate/2]).
 
+-type dirty_constraint_result() :: true | {true, any()} | false.
+-type constraint_result() :: true | {true, any()} | {false, fun() | atom()}.
 -type constraint() :: int | nonempty | fun().
--export_type([constraint/0]).
+-export_type([constraint/0, constraint_result/0]).
 
--spec validate(binary(), [constraint()]) -> true | {true, any()} | false.
+-spec validate(binary(), [constraint()]) -> constraint_result().
 validate(Value, [Constraint]) ->
 	apply_constraint(Value, Constraint);
 validate(Value, Constraints) when is_list(Constraints) ->
@@ -37,17 +39,21 @@ validate_list(Value, [Constraint|Tail], State) ->
 			validate_list(Value, Tail, State);
 		{true, Value2} ->
 			validate_list(Value2, Tail, modified);
-		false ->
-			false
+		Else ->
+			Else
 	end.
 
 %% @todo {int, From, To}, etc.
-apply_constraint(Value, int) ->
-	int(Value);
-apply_constraint(Value, nonempty) ->
-	nonempty(Value);
+apply_constraint(Value, int = F) ->
+	prepare_result(int(Value), F);
+apply_constraint(Value, nonempty = F) ->
+	prepare_result(nonempty(Value), F);
 apply_constraint(Value, F) when is_function(F) ->
-	F(Value).
+	prepare_result(F(Value), F).
+
+-spec prepare_result(dirty_constraint_result(), fun() | atom()) -> constraint_result().
+prepare_result(False, F) when False =:= false -> {false, F};
+prepare_result(True, _F)                      -> True.
 
 %% Constraint functions.
 
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -1253,7 +1253,10 @@ kvlist_to_map(Keys, [{Key, Value}|Tail], Map) ->
 filter([], Map) ->
 	Map;
 filter([{Key, Constraints}|Tail], Map) ->
-	filter_constraints(Tail, Map, Key, maps:get(Key, Map), Constraints);
+	case maps:find(Key, Map) of
+		{ok, Val} -> filter_constraints(Tail, Map, Key, Val, Constraints);
+		_         -> throw({bad_constraint, Key})
+	end;
 filter([{Key, Constraints, Default}|Tail], Map) ->
 	case maps:find(Key, Map) of
 		{ok, Value} ->
@@ -1270,7 +1273,9 @@ filter_constraints(Tail, Map, Key, Value, Constraints) ->
 		true ->
 			filter(Tail, Map);
 		{true, Value2} ->
-			filter(Tail, maps:put(Key, Value2, Map))
+			filter(Tail, maps:put(Key, Value2, Map));
+		{false, F} ->
+			throw({bad_constraint, {Key, F}})
 	end.
 
 %% Tests.
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -1250,32 +1250,40 @@ kvlist_to_map(Keys, [{Key, Value}|Tail], Map) ->
 %% Loop through fields, if value is missing and no default, crash;
 %% else if value is missing and has a default, set default;
 %% otherwise apply constraints. If constraint fails, crash.
-filter([], Map) ->
-	Map;
-filter([{Key, Constraints}|Tail], Map) ->
-	case maps:find(Key, Map) of
-		{ok, Val} -> filter_constraints(Tail, Map, Key, Val, Constraints);
-		_         -> throw({bad_constraint, Key})
-	end;
-filter([{Key, Constraints, Default}|Tail], Map) ->
-	case maps:find(Key, Map) of
-		{ok, Value} ->
-			filter_constraints(Tail, Map, Key, Value, Constraints);
-		error ->
-			filter(Tail, maps:put(Key, Default, Map))
-	end;
-filter([Key|Tail], Map) ->
-	true = maps:is_key(Key, Map),
-	filter(Tail, Map).
+filter(Fields, InputM) ->
+	filter(Fields, InputM, #{}).
 
-filter_constraints(Tail, Map, Key, Value, Constraints) ->
+filter([], InputM, ErrM) ->
+	case maps:size(ErrM) of
+		0 -> InputM;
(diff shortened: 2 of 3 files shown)


tuscland:
Hi,

I would like to mention that this "constraints" problem is not limited to Cowboy route constraints but also other types of input validation and transformation.

I have written a small module that uses function composition (think "type composition") and validate input the same way Cowboy does at the moment, that is, return `{true, NewValue}` or `{false, Reason}` where `Reason` is a programmatic description of the error (an erlang term).

This code is currently used for JSON or XML input as well. I use it with Cowboy constraints too, by waiving the `Reason` away, and returning `false` instead.

If Cowboy provided a callback function to allow for error processing, then route constraints would not be to be more complex than to return `{false, Reason}` when a constraint fails. The custom error handler would then take care of doing what is needed (e.g. format the error, set the body and reply the proper error code).

Camille

essen:
The subject of constraints has come up again in the context of reverse routing: #816 

At this point I want to do a few more interface changes and tag 2.0.0-pre.2. I don't think constraints will be solved yet. This is a much larger problem than I thought.

In summary, so far, constraints need:

* To be able to convert data to a usable type.
* To have a normal and inverse operation `int(binary()) -> integer() and inverse_int(integer()) -> binary()` to make reverse routing possible.
* To return `{ok, Value}` or `{error, Reason}`. Indeed this makes more sense.
* The `Reason` should probably be an atom.
* For the cases where the error is not catched in user code we should probably have a hook. This hook can be used to do formatting of the error and send an appropriate response. Otherwise Cowboy would just send a 400 with its own formatting.

I believe a hook is better because then you don't have to worry about it per handler, it will be used anywhere in the listener.

Am I missing something?

tuscland:
Hello Loïc,

Please let me contribute my experience on that matter.

We have developped a simple module that we use to convert data into usable type for this very purpose. It is compatible with Cowboy's constraints, and we use it also to parse and validate JSON values. I'd be happy to send this code to you if you want to gather ideas.

It uses function composition to express the parsing intent, e.g. if you evaluate `list(integer())`, or `any(null(), uuid)`, you get a function that can be used to parse binary input, and the result of calling this function is `{true, Result}` or `{false, Reason}`.

So in our system, we have a description of the structure and hierarchy of the URLs. This description is compiled to a list of strings that are used by Cowboy for routing, and to a list of functions that we use to generate URLs (reverse-routing if you will). This makes the routes definition practically empty since it is auto-generated.

Best,
Camille

essen:
This is interesting to get inspiration from when it comes to providing default constraints: http://livr-spec.org/

essen:
I've done both this and reverse constraints in c22173037134becbac882533f80339f4127a8ad2. Feedback welcome!

manifest:
it would be useful to access not only a value but also its key in case of `format_error`.

essen:
This module only validates values. I can see the exit exception having the field though, there it would definitely make sense (and it probably would make even more sense to aggregate all errors and only throw at the end).

essen:
Something like this?

``` diff
diff --git a/src/cowboy_req.erl b/src/cowboy_req.erl
index 3eca379..f65ecc6 100644
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -762,28 +762,43 @@ kvlist_to_map(Keys, [{Key, Value}|Tail], Map) ->
 		kvlist_to_map(Keys, Tail, Map)
 	end.
 
+filter(Fields, Map0) ->
+	case filter(Fields, Map0, #{}) of
+		{ok, Map} ->
+			Map;
+		{error, Errors} ->
+			exit({validation_failed, Errors})
+	end.
+
 %% Loop through fields, if value is missing and no default, crash;
 %% else if value is missing and has a default, set default;
 %% otherwise apply constraints. If constraint fails, crash.
-filter([], Map) ->
-	Map;
-filter([{Key, Constraints}|Tail], Map) ->
-	filter_constraints(Tail, Map, Key, maps:get(Key, Map), Constraints);
-filter([{Key, Constraints, Default}|Tail], Map) ->
+filter([], Map, Errors) ->
+	case maps:size(Errors) of
+		0 -> {ok, Map};
+		_ -> {error, Errors}
+	end;
+filter([{Key, Constraints}|Tail], Map, Errors) ->
+	filter_constraints(Tail, Map, Errors, Key, maps:get(Key, Map), Constraints);
+filter([{Key, Constraints, Default}|Tail], Map, Errors) ->
 	case maps:find(Key, Map) of
 		{ok, Value} ->
-			filter_constraints(Tail, Map, Key, Value, Constraints);
+			filter_constraints(Tail, Map, Errors, Key, Value, Constraints);
 		error ->
-			filter(Tail, Map#{Key => Default})
+			filter(Tail, Map#{Key => Default}, Errors)
 	end;
-filter([Key|Tail], Map) ->
-	true = maps:is_key(Key, Map),
-	filter(Tail, Map).
+filter([Key|Tail], Map, Errors) ->
+	case maps:is_key(Key, Map) of
+		true ->
+			filter(Tail, Map, Errors);
+		false ->
+			filter(Tail, Map, Errors#{Key => required})
+	end.
 
-filter_constraints(Tail, Map, Key, Value0, Constraints) ->
+filter_constraints(Tail, Map, Errors, Key, Value0, Constraints) ->
 	case cowboy_constraints:validate(Value0, Constraints) of
 		{ok, Value} ->
-			filter(Tail, Map#{Key => Value});
+			filter(Tail, Map#{Key => Value}, Errors);
 		{error, Reason} ->
-			exit(Reason)
+			filter(Tail, Map, Errors#{Key => Reason})
 	end.
```

essen:
Well I pushed in most recent commit. :-)

manifest:
I see. So, I've got

```erlang
{error, Constrains} = cowboy_req:match_qs(...)
```

Next, I could execute

```erlang
maps:map(fun(Key, Val) -> fun cowboy_constraints:format_error(Val) end, Constrains)
```

Does the `map` is a right container then?

essen:
Depends on what you're going to do with it. Would certainly be better if you were going to put errors next to form fields in an HTML page for example. And conversion from this map should be fast if you need a list (it's unlikely to be a large map).

PR #780, opened 2015-01-14
Fix reply/4 spec

egobrain:
There is wrong -spec of reply/4 function 3rd parametr type must be same as [#http_req spec](https://github.com/ninenines/cowboy/blob/master/src/cowboy_req.erl#L140)

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -720,7 +720,8 @@ reply(Status, Headers, Req=#http_req{resp_body=Body}) ->
 	reply(Status, Headers, Body, Req).
 
 -spec reply(cowboy:http_status(), cowboy:http_headers(),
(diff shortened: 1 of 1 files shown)


essen:
Do you have an example code that makes Dialyzer fail?

kpy3:
Code that fail:

process(Fun, Req, State) ->
    Req1 = cowboy_req:set_resp_body_fun(chunked, Fun, Req),
    {ok, Req2} = cowboy_req:reply(200, Req1),
    {ok, Req2, State}.

Dialyzer in Erlang R16B03-1 found this code correct, but dialyzer from Erlang 17.4 not.

essen:
OK. Can you do a tiny change and move the | at the beginning of the second line instead of the end of the first line? Thanks.

egobrain:
Fixed

essen:
Merged, thanks!

PR #785, opened 2015-01-30
Reply with 400 on if*-match parsing crash

martinrehfeld:
Requesting a `cowboy_rest` resource previously crashed the connection process when the value of the `If-Match` or `If-None-Match` headers did not include a quoted string as expected, e.g.
```
curl -v -H 'If-None-Match: X' ...
```
would crash.

This PR changes the behaviour to return a `400 Bad Request` instead.

Please let me know if anything needs to be touched up before it can be merged.

Cheers,
Martin

### DIFF
--- a/src/cowboy_http.erl
+++ b/src/cowboy_http.erl
@@ -668,7 +668,9 @@ token(<< C, Rest/binary >>, Fun, Case, Acc) ->
 
 -spec quoted_string(binary(), fun()) -> any().
 quoted_string(<< $", Rest/binary >>, Fun) ->
-	quoted_string(Rest, Fun, <<>>).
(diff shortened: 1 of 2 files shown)


essen:
OK but why add a try/catch there if we already handle it?

martinrehfeld:
I just added that in case any other crash would occur. But you are right, it is not needed to fix the specific problem. Shall I remove it?

essen:
Yes please.

martinrehfeld:
Done.

essen:
Thanks. Should be good but I'll test and merge after I settle down next week.

martinrehfeld:
Sure. Thanks for your feedback.

essen:
Merged, thanks!

PR #808, opened 2015-03-26
fixed cowboy_clock inbox overflow if system clock was changed

asskovalev:
For some reasons I have disabled compensation of system time in my application (http://www.erlang.org/doc/man/erl.html#+c) and realized abnormal memory consumption.

If system starts with a time lag and after a period of time synchronizes system clock to NTP, cowboy_clock receives a huge number of 'update' messages. If the lag is large (e.g. a year), node stucks.

To fix this situation I've changed timer:send_interval to erlang:send_after

### DIFF
--- a/src/cowboy_clock.erl
+++ b/src/cowboy_clock.erl
@@ -84,6 +84,7 @@ handle_cast(_Msg, State) ->
 
 -spec handle_info(any(), State) -> {noreply, State} when State::#state{}.
 handle_info(update, #state{universaltime=Prev, rfc1123=B1, tref=TRef0}) ->
+	%% Cancel the timer in case an external process sent an update message.
 	erlang:cancel_timer(TRef0),
 	T = erlang:universaltime(),
 	B2 = update_rfc1123(B1, Prev, T),
--- a/src/cowboy_clock.erl
+++ b/src/cowboy_clock.erl
@@ -65,7 +65,7 @@ init([]) ->
 		named_table, {read_concurrency, true}]),
 	T = erlang:universaltime(),
 	B = update_rfc1123(<<>>, undefined, T),
-	{ok, TRef} = timer:send_interval(1000, update),
+	TRef = erlang:send_after(1000, self(), update),
 	ets:insert(?MODULE, {rfc1123, B}),
 	{ok, #state{universaltime=T, rfc1123=B, tref=TRef}}.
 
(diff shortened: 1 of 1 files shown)


maxlapshin:
@essen  this fix seems to be rather important for us, because cowboy_clock is the only process inside flussonic that gets mad on NTP clock adjust.

essen:
Wow that's a weird one. :-)

src/cowboy_clock.erl
```diff
@@ -83,10 +83,12 @@ handle_cast(_Msg, State) ->
 	{noreply, State}.
 
 -spec handle_info(any(), State) -> {noreply, State} when State::#state{}.
-handle_info(update, #state{universaltime=Prev, rfc1123=B1, tref=TRef}) ->
+handle_info(update, #state{universaltime=Prev, rfc1123=B1, tref=TRef0}) ->
+	erlang:cancel_timer(TRef0),
```
essen: You shouldn't need to cancel the timer after the message has been delivered. That function call probably does nothing and just returns false, unless I missed something.

maxlapshin:
Yes. If you do not rely on some specific timer:send_interval behaviour, then this patch fixes a situtation that happens with one year jump in time.

src/cowboy_clock.erl
maxlapshin: I think that you are wrong.  If I send another message  update to this cowboy_clock, I will create a second wave of timers: there will exist two timers that will send update to this process and they will leave in different waves.

This line is very important because it removes appearing of second wave of timers.

essen:
```
Cancels a timer, where TimerRef was returned by either erlang:send_after/3 or erlang:start_timer/3. If the timer is there to be removed, the function returns the time in milliseconds left until the timer would have expired, otherwise false (which means that TimerRef was never a timer, that it has already been cancelled, or that it has already delivered its message).
```

I believe it has already delivered its message so there is nothing to cancel. Easy to verify, just check the return value.

maxlapshin: I mean the situation when "update" is sent second time from outer source.

Well, this is just a common protection pattern that we are using everywhere in our code.

essen: I do not understand what you mean by "second time" and "outer source" then.

maxlapshin: What will happen with erlang:send_after if I make from my code: whereis(cowboy_clock) ! update ?

It will receive update message, create new timer, not cancel old timer and will get 2 update messages during next second.

This will happen only if I send update message from my part of code, of course.

essen: Right. Wondering though, do you need to be able to send the message?

maxlapshin: No, I don't need to =)

As I've told, this pattern is just our common pattern that we are using everywhere.

essen: Yep. Then I'm fine with it but would like a comment above the send_after like

```
%% Cancel the timer in case an external process sent an update message.
```

Otherwise I risk removing it by mistake in the future.

asskovalev: Done

essen:
One doubt remains though. The docs for cancel_timer say that the message may already be in the mailbox when you call the function, so we would need to have a `receive update -> ok after 0 -> ok end` after the cancel_timer call to be completely safe.

maxlapshin:
Yes. Most safe is a small loop that extracts all update messages, but just cancelling timer is ok, because it stops new wave of sending messages.

essen:
Disregard what I said last I was incorrect.

essen:
Merged in 1.0.x and master. Cheers!

PR #818, opened 2015-05-06
Exit with exit/1 on handler exception and include class in reason

fishcakez:
Using `exit/1` will prevent emulator messages being generated when a handler raises an error or throw. It is also a precursor to supporting `proc_lib` as the class would be lost when using `proc_lib`.

Would it be better to tag the exit reasons, such as `{handler_exception, [{class, ..},..]}`?

### DIFF
--- a/src/cowboy_handler.erl
+++ b/src/cowboy_handler.erl
@@ -52,13 +52,14 @@ execute(Req, Env) ->
 		Stacktrace = erlang:get_stacktrace(),
 		cowboy_req:maybe_reply(Stacktrace, Req),
 		terminate({crash, Class, Reason}, Req, HandlerOpts, Handler),
-		erlang:Class([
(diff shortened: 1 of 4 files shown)


essen:
What use do you think of for tagging?

fishcakez:
For pattern matching on in the `error_logger`.

essen:
Fair enough.

But then I think a name more explicit that this is an error in cowboy user code is in order. How about {cowboy_handler, [...]}?

fishcakez:
Sure.

essen:
Merged, thanks!

Feel free to make a proposal before sending the next PR so I can coordinate.

fishcakez:
I will just do handling of parent exit signal messages next. However after that coordination will be required.

essen:
Cool.

PR #805, opened 2015-03-25
Get stacktrace before executing rest_terminate

danielwhite:
At present, the resulting stacktrace will be misleading about where the
underlying error occurred.

This has been resolved in the 2.x branch already, due to some other
changes, but is surprising in existing deployments.

PR #823, opened 2015-05-15
Save stacktrace sooner in cowboy_rest

RumataEstor:
This prevents stacktrace information be messed up by the calls inside rest_terminate.


essen:
Already have #805 covering it. Will get to it on Monday or earlier. Thanks! :-)

RumataEstor:
I see #805 is still open...

This is a serious problem that hides the real culprit of an error and makes a search for it impossible.

essen:
Yep, being real busy preparing things for the EUC next week. Will do a round of merges after that, or if lucky a little before. Meanwhile people can cherry-pick commits from PR if they need.

PR #828, opened 2015-05-29
expect header parses to an atom (continue)

jmccaughey:
Cowboy is awesome -- thanks for making it!  Maybe the switch to cow_lib for parsing headers on Feb 4 caused this?  

My app was latent by one or two seconds every post request because the client was doing Expect: Continue but not getting the "go ahead" back


essen:
Good catch, thanks!

essen:
Merged!

PR #832, opened 2015-06-12
Add support for sanitizing user provided input

akshaykodali:
Enforces Rule #1
https://www.owasp.org/index.php/XSS_%28Cross_Site_Scripting%29_Prevention_Cheat_Sheet

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -521,6 +521,30 @@ body_qs(Req, Opts) ->
 			{badlength, Req2}
 	end.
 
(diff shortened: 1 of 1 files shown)


essen:
Hi, Cowboy does not deal with HTML, it's only an HTTP server.

This kind of code would fit in an HTML sanitization or a template library. I don't know one (other than erlydtl, dtl or mustache template libraries) so perhaps you could start one?

Thanks!

akshaykodali:
My bad, essen. ErlyDTL has built in support for escaping user provided HTML. Sorry for the trouble.

PR #845, opened 2015-07-05
Use HTTPS dependencies when using Rebar

ahf:
HTTPS gives encryption and authentication, which the Git protocol lacks. Use it for Rebar builds as well :-)


essen:
Merged, thanks!

PR #850, opened 2015-07-15
Fix typo in cowboy_http:asctime_date

mocchira:
`{error, badarg1}` seems to be wrong.
https://github.com/ninenines/cowboy/blob/1.0.x/src/cowboy_http.erl#L428
should be replaced with `{error, badarg}`


essen:
Not a typo, that's actually how it used to be, and it was left over from newer changes.

Thanks, I'll merge it probably tomorrow, I need to get 1.0.x onboard the CI train first.

essen:
Merged, thanks!

PR #862, opened 2015-07-28
Use 'Connection' header when necessary

### LINKED ISSUE #839: apache benchmark (ab) using keepalive is not working
I'm using the last code from master.

The app is very simple (based on hello world rest example)

```erlang
-module(toppage_handler).

-export([init/2, start/0]).
-export([content_types_provided/2]).
-export([hello_to_html/2]).

init(Req, Opts) ->
    {cowboy_rest, Req, Opts}.

content_types_provided(Req, State) ->
    {[
        {<<"text/html">>, hello_to_html}
    ], Req, State}.

hello_to_html(Req, State) ->
    {<<"Hello world">>, Req, State}.

start()->

    ok = application:start(crypto),
    ok = application:start(ranch),
    ok = application:start(cowlib),
    ok = application:start(cowboy),

    Dispatch = cowboy_router:compile(
        [
            {'_', [{"/", toppage_handler, []}]}
        ]),

    {ok, _} = cowboy:start_http(http, 100, [{port, 8080}], [{env, [{dispatch, Dispatch}]}]).
```

I'm trying to benchmark using:

ab -n 100000 -c 1500 -k  http://172.17.4.107:8080/

If -k is present in the arguments nothing happened. Without this one is working fine. I tested on a other webservers the command and is working properly.

Any idea what I'm doing wrong?

Silviu

binarin:
Is this a right way to fix the issue? If it's OK, I'm going to move this logic about 'Connection' to small helper function and to add some tests.

Fixes #839 when 'Connection: Keep-Alive' wasn't sent in a HTTP/1.0
response. Now the usage of 'Connection' header is consistent with
current protocol version: when this header is not specified explicitly
in the response, HTTP/1.0 implies 'Connection: close' and HTTP/1.1
implies 'Connection: Keep-Alive'. So if current 'Connection' value
matches the default value of current protocol, we won't state obvious
fact in the response; and vice versa.

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -685,8 +685,10 @@ reply(Status, Headers, Body, Req=#http_req{
 		method=Method, resp_compress=Compress,
 		resp_state=RespState, resp_headers=RespHeaders})
 		when RespState =:= waiting; RespState =:= waiting_stream ->
-	HTTP11Headers = if
-		Transport =/= cowboy_spdy, Version =:= 'HTTP/1.1' ->
+	StdHeaders = if
+		Transport =/= cowboy_spdy, Version =:= 'HTTP/1.0', Connection =:= keepalive ->
+			[{<<"connection">>, atom_to_connection(Connection)}];
+		Transport =/= cowboy_spdy, Version =:= 'HTTP/1.1', Connection =:= close ->
 			[{<<"connection">>, atom_to_connection(Connection)}];
 		true ->
 			[]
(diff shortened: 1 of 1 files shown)


essen:
I'm not sure. I think the usage of Connection for HTTP/1.0 should not be enabled by default. https://tools.ietf.org/html/rfc7230#appendix-A.1.2 points out that the results will be more or less random depending on how the client implemented it. The real fix that I can accept *right now* is to close the connection instead of keeping it open (which was what Cowboy did before the issue was introduced).

essen:
You also broke 16 tests here: https://circle-artifacts.com/gh/ninenines/cowboy/46/artifacts/0/home/ubuntu/cowboy/logs/ct_run.ct@localhost.2015-07-28_20.45.33/ubuntu.cowboy.http_SUITE.logs/run.2015-07-28_20.45.35/suite.log.html

essen:
OK this is an

essen:
OK this is an issue that I can easily observe, the fix looks reasonable, I will just have to make sure the tests don't break. Will do that now.

binarin:
After reading the RFCs appendix that you've mentioned, I agree that Keep-Alive should be disabled in HTTP/1.0 altogether. But I don't have any time to make that patch in the nearest future.

essen:
Yep I'm adding tests for the header right now. Will make that an option later on, it's unrelated.

essen:
OK I pushed your commit amended to master. I will backport it to 1.0.3 this afternoon.

Thanks!

PR #863, opened 2015-07-29
fix broken rest_basic_auth example

brigadier:


essen:
I'm not sure if I merged something or fixed myself but the code has this now. Thanks!

PR #865, opened 2015-07-30
Correct handling of missing echo parameter.

technion:
In the example echo GET server, the "undefined" pattern match on line 14 will never be reached. This is demonstrated in this gist:

https://gist.github.com/technion/daffb546d3454ae84ae3

Here we see that, if the echo parameter is not defined in the query, we get a badarg crash on line 10. This PR provides the most simple solution. By defining a default, the undefined match works as expected.

It may be worth considering a bigger picture that perhaps this should be default - it would be unusual to be preferable for an end user to trigger a badarg crash, as opposed to handling it as per the example. The code in this PR, imo, looks a lot more ugly than the code it replaced, and end users are likely to choose the simpler option of ignoring it.

### DIFF
--- a/examples/echo_get/src/toppage_handler.erl
+++ b/examples/echo_get/src/toppage_handler.erl
@@ -7,7 +7,7 @@
 
 init(Req, Opts) ->
 	Method = cowboy_req:method(Req),
-	#{echo := Echo} = cowboy_req:match_qs([echo], Req),
+	#{echo := Echo} = cowboy_req:match_qs([{echo, [], undefined}], Req),
 	Req2 = echo(Method, Echo, Req),
 	{ok, Req2, Opts}.
 


technion:
As far as I can see, this commit never touched those failing tests?

essen:
We do want to crash and the undefined clause is a leftover that needs to be removed.

What is missing though is the silencing of errors occuring in input validation, and perhaps returning 400.

-- 
Loïc Hoguin
http://ninenines.eu

-------- Original Message --------

technion:
>silencing of errors occuring in input validation and perhaps returning 400

Well yes, that was my general goal with catching it, but if there are better plans for doing so I'm happy to abandon this PR.

technion:
I can see that the bigger issue is likely to be addressed in #776 so I'll close this off and get it out of your way. For reference, I've been quite happy testing out this pattern in pre-2, being:

* Default each variable to unknown
* Use maps:filter to identify any keys with a "undefined" value
* Return 400 based on the output of that if required

PR #869, opened 2015-08-05
Update rebar.config to reflect erlang.mk

jlouis:
Rather simple patch that tries to bring master into order with what erlang.mk requires in versions.


essen:
I know there's warnings, but I can't do anything about that. Soon that file will be entirely generated, and I don't have that kind of information to fill it in. So merging this now to remove it a couple weeks later doesn't make much sense.

essen:
It's now in sync and will stay in sync because the file is now generated. Thanks!

Issue #884, opened 2015-09-05
 Ranch listener http terminated


glejeune:
In certain condition, the ranch listener crash with error :

```
Ranch listener http had connection process started with cowboy_protocol:start_link/4 at <0.160.0> exit with reason: {function_clause,[{cow_http_hd,etag_list,[<<"AAAA">>,[]],[{file,"src/cow_http_hd.erl"},{line,1894}]},{cow_http_hd,parse_if_none_match,1,[{file,"src/cow_http_hd.erl"},{line,1958}]},{cowboy_rest,if_none_match_exists,2,[{file,"src/cowboy_rest.erl"},{line,725}]},{cowboy_protocol,execute,4,[{file,"src/cowboy_protocol.erl"},{line,470}]}]}
```

So, this happen when the `If-None-Match` is mal formated, so when the value is `xyz` and not `"xyz"`.

You can reproduce by launching the `rest_hello_world` example and by using this curl command :

```
curl -v -H "If-None-Match: AAAA" -X GET http://localhost:8080
```

essen:
This is on master or 1.0?

glejeune:
Oups, sorry... On master

essen:
Thanks. I think this is something that popped up before but I can't find the ticket now. Will also need to check 1.0 for this.

glejeune:
Same problem with 1.0.

JoshRagem:
just ran into this myself. if the etag header value is missing one or both of the surrounding quotes this error happens:

```
[...]

PR #885, opened 2015-09-06
Issue #884: Ranch listener http terminated (master)

glejeune:
Malformated etag with If-Match and If-None-Match are ignored


essen:
This has resulted in a 400 since Cowboy 2.0. I think this is the right way to go since we can't evaluate the precondition and determine if it's true (continue) or false (412).

Closing, thanks!

PR #886, opened 2015-09-06
Issue #884: Ranch listener http terminated (1.0.x)

glejeune:
Malformated etag with If-Match and If-None-Match are ignored

### DIFF
--- a/src/cowboy_rest.erl
+++ b/src/cowboy_rest.erl
@@ -523,7 +523,7 @@ resource_exists(Req, State) ->
 
 if_match_exists(Req, State) ->
 	State2 = State#state{exists=true},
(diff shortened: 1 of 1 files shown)


essen:
Looks fine, thanks!

essen:
Hm it will need some test though, otherwise it risks breaking again.

essen:
Also probably better to base this off master as I prefer to backport things than the opposite.

essen:
Can you tell me why it should be ignored rather than result in an error? If-Match is a pretty important header and I can't find anything in the RFCs that say we should ignore them. If it's to fix an erroneous client, can you add a comment saying which client/version it is and under which conditions it happens?

essen:
Closing this one, leaving the PR against master for now. #885 Thanks!

Issue #887, opened 2015-09-06
query string parsing fails with error 500 if a query parameter is empty


galdor:
The documentation for cowboy HEAD [indicates](http://ninenines.eu/docs/en/cowboy/HEAD/guide/req/):

> If no default is provided and the value is missing, the query string is deemed invalid and the
> process will crash.

In practice, the handler crashes and the client receives a 500 error. But it really should be a 400 error.

essen:
Agree on both counts, and I confirm it was parsing those properly before. Can you provide a crash log?

galdor:
Sorry, I just modified my message. The problem is not that empty query parameters are not supported. It's that the response is 500 and not 400 (something is wrong with the query, it's not an internal error).

essen:
Alright. Should be an easy patch, check do_maybe_reply in cowboy_req if you want to send a PR. Thanks!

essen:
For future reference: The patch in the closed PR above is good, just incomplete (lacking tests).

PR #888, opened 2015-09-06
reply with a 400 status code when query string parsing fails

galdor:
See #887.

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -872,6 +872,12 @@ do_maybe_reply([{erlang, binary_to_integer, _, _}, {cow_http_hd, parse_content_l
 	cowboy_req:reply(400, Req);
 do_maybe_reply([{cow_http_hd, _, _, _}|_], Req) ->
 	cowboy_req:reply(400, Req);
+do_maybe_reply([{cowboy_constraints, _, _, _}|_], Req) ->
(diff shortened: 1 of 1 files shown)


galdor:
Using a stack trace to determine which status code to send seems incredibly fragile to me. Is there any plan to return errors differently ? Note that I am a complete erlang beginner, so I may be wrong.

essen:
Hm you add three things, what are they? Which one is the one you did encounter?

Fragile yes, which is why this will need tests. But not much choice if we want to keep the "let it crash" behavior.

galdor:
There are three possibles crashes:

- The parameter is missing.
- The parameter has an empty value.
- The parameter has an invalid value.

The problem here is having to match the stacktraces of all possible errors everywhere in a single function. "Let it crash" here means not being able to send the right status code at the right place.

The other problem is that even with a 400 status code, there is no indication of what is wrong with the request. So if you have a public API, and a client sends a wrong request, he gets a response without any error message. This is a very real problem in general.

essen:
This function is there so we can follow the HTTP spec, not to make people's life easy. :-)

About knowing what's wrong, there's a ticket somewhere about improving the constraints to have details about all the failures in the requested input. Once we have that we can easily add an optional error_logger call alongside sending 400 to explain what went wrong.

PR #899, opened 2015-10-05
add interface for custom protocol

comtihon:
Add interface for setting transport and protocol custom modules (for example to solve proxy protocol issues via ranch_proxy or custom cowboy_protocol parsing mechanics).


essen:
You can call ranch:start_listener/6 for these cases.

comtihon:
@essen but in cowboy it is most comfortable and also - it should be done so, because if there appears some other logic in the start_http interface - we should maintain it in our logic, where we call ranch.

essen:
These are convenience functions, that's all they'll ever be. In 2.0 they're being replaced with start_tcp and start_tls, but they're still just a tiny alias to start_listener. Adding more functions (this, or child_spec directly in Cowboy) is just unnecessary.

comtihon:
ok

PR #901, opened 2015-10-18
Move ranch dep to 1.1.0, fixes build

febeling:
The rebar.config wasn't in sync with Makefile deps and compilation was failing. This fixes it.


essen:
It's now in sync and will stay in sync because the file is now generated. Thanks!

PR #907, opened 2015-10-28
Update README.md

Tombert:
I noticed that this appeared to point to a bad URL.  If LeoFS is sponsoring this, they'd probably prefer having the right URL.


essen:
Merged, thanks!

Issue #871, opened 2015-08-07
cowboy websocket wrong behavior on stop in init.


comtihon:
In my websocket handler's `Module:init/2` I need to stop handler in some clause.
I use `{stop, Req, undefined};` as in the [documentation](http://ninenines.eu/docs/en/cowboy/HEAD/guide/ws_handlers/).
But instead stopping handler - I got an error `{undef,[{stop,upgrade..`. So, as I can see - instead stopping the process cowboy tries to call `stop:upgrade`, threating stop as a callback Module.

essen:
Yep it looks like the documentation is outdated on this part. The correct behavior is to return ok, I believe, though that might change in the future. Note that you need to send yourself a reply if you want to send something other than 200.

comtihon:
fix in https://github.com/ninenines/cowboy/pull/912

PR #912, opened 2015-11-06
made it ranch_proxy_protocol compatible

comtihon:
Fixes https://github.com/ninenines/cowboy/issues/871 and https://github.com/heroku/ranch_proxy_protocol/issues/16
The thing is, that custom socket record (which ranch does) is not matched in cowboy_websocket, as it receives socket as port.


src/cowboy_websocket.erl
```diff
@@ -411,3 +412,10 @@ handler_terminate(#state{env=Env, handler=Handler},
 		Req, HandlerState, Reason) ->
 	cowboy_handler:terminate(Reason, Req, HandlerState, Handler),
 	{ok, Req, [{result, closed}|Env]}.
+	
+	
+%% @private
+get_real_socket({proxy_socket, _, Socket, _, _, _, _, _, _}) -> Socket;
```
ferd: This is really not a great idea. We could be changing the structure any time.

The proper way to call it for us is with `Transport:match_port(Sock)` (as in https://github.com/heroku/ranch_proxy_protocol/blob/master/src/ranch_proxy_protocol.erl#L261-L262) -- this is non-standard and our own internal work-around, and we haven't done the due process of discussing with @essen as if that would be in any way a proper way to go about things.

So I figure that before pulling support in cowboy for this, we should check with @essen if he feels things should be done differently or if he at least agrees with our approach of allowing specific 'match ports', and also avoid internal matching like this.

comtihon: Yes, I undertand you and absolutely agree, but this is a quick fix.

essen: I am still a bit too busy to get back to work just yet (and, to be frank, this is a topic that I will need time to get to in general, although I understand the problem with socket matching), but regardless this kind of patch will never be merged for the reasons @ferd mentioned.

My guts tell me extending the messages() function to return the socket part to match against is the right solution, *but* I don't think it's a good idea to take a decision on this until I get some practical experience with this (writing an example and experimenting things out) and until the changes to the HTTP/1.1 protocol get pushed, as they have more similar receive clauses and it won't be just Websocket anymore like this. And, similarly, the code to hook before Cowboy starts parsing HTTP will also be different. And then there's also the question of testing things out and such. It's such a large task that I didn't plan this for Cowboy 2 (CONNECT is in, but PROXY is still out) and while I don't rule out having some support in Cowboy 2, I believe we will need more time before PROXY support (or code to enable PROXY support and other) becomes stable.

Thoughts on messages(Socket) -> {OK, Closed, Error, MatchSocket}?

ferd: expanding `messages/1` would probably work fine for us, it just makes a progressive deployment difficult if the support isn't baked in ahead of time. It's workable, though.

essen: It's a new function, messages/0 becomes messages/1 (to have the socket in arguments), and you can still keep a messages/0 around in your deployment for a while to ease transition.

Will link to this thread from the Ranch 2 meta ticket.

ferd: Ah yeah that sounds entirely workable.

AeroNotix:
The whitespace changes make this PR more work to read than is necessary.

AeroNotix:
Having support for this is very useful when using WebSockets on AWS ELBs.

essen:
In its current form it has no chance of being merged. It's a hack, and one without tests either.

I'm open to proper solutions though, perhaps extending the Transport:messages() function to also return the real socket.

comtihon:
Sorry, @essen , I have no time for it now. I closed it as I thought that is was fixed in cowboy master branch, but forgot to use proxy protocol, so I reopened it.
I completely agree with you, as soon as I will have time I'll think how to make this gracefully.

essen:
Yep no worries, it's good to have the ticket around to keep this in mind either way.

essen:
Closing, I'm still not against adding some support but I need tests and docs at the very least. Thanks!

jimdigriz:
If anyone is curious, I needed proxy protocol support in cowboy2 so rolled up my sleeves: https://github.com/ninenines/cowboy/compare/master...jimdigriz:proxy

Changes [`ranch_proxy_protocol`](https://github.com/heroku/ranch_proxy_protocol) side are: https://github.com/heroku/ranch_proxy_protocol/compare/master...jimdigriz:cowboy

PR #919, opened 2015-11-28
minor rebar.config style fix

danieldreier:
Update rebar.config to use {branch, Name} / {tag, Tag} syntax. rebar3 produces a warning without this change. Does not break rebar 1 compatibility.


essen:
rebar.config is generated automatically, and this information is unavailable in Erlang.mk.

danieldreier:
makes sense, thanks for the quick response!

PR #925, opened 2015-12-14
make a failed constraint return a 400 bad request code

### LINKED ISSUE #923: Return 400 error code on constraint failure
Hi,

For now, if I'm not mistaken, cowboy can only return a 500 Server Error when a queries constraint fails (returns `false`).
It would be great if we could customize that error code. For example I would prefer a 400 Bad Request.

Would you be okay for a PR with this? How would you want this to be implemented?

ibizaman:
Add a clause matching when a constraint fails and which returns a 400 bad request code instead of the 500 server error.

fixes #923


essen:
Looks fine, but does miss a test.

ibizaman:
Yes exactly, but I was wondering where to put it. I'm fairly new to erlang and even more to erlang testing.
Would that be in the http_SUITE.erl file?

essen:
Yes. Check the test against `/crash/content-length`. Then add a new clause to `input_crash_h`. This handler and test is used to make sure user input results in a 400.

essen:
I've just ensured this is done properly, the code has changed a lot so the solution is fairly different now. Closing, thanks!

PR #928, opened 2015-12-22
Allow colon within path segments

fp:
Allow `cowboy_router:compile` to handle colon characters within
path segments, rather than exiting with `badarg`.  This is allowed
via RFC 7230 2.7 -> [RFC 3986 3.3](https://tools.ietf.org/html/rfc3986#section-3.3):

```
      segment       = *pchar
      segment-nz    = 1*pchar
      segment-nz-nc = 1*( unreserved / pct-encoded / sub-delims / "@" )
                    ; non-zero-length segment without any colon ":"

      pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```

### DIFF
--- a/src/cowboy_router.erl
+++ b/src/cowboy_router.erl
@@ -99,12 +99,11 @@ compile_rules(<< S, Rest/bits >>, S, Segments, Rules, <<>>) ->
 	compile_rules(Rest, S, Segments, Rules, <<>>);
 compile_rules(<< S, Rest/bits >>, S, Segments, Rules, Acc) ->
 	compile_rules(Rest, S, [Acc|Segments], Rules, <<>>);
+%% only colon on segment start is special, otherwise allow in path segment
 compile_rules(<< $:, Rest/bits >>, S, Segments, Rules, <<>>) ->
 	{NameBin, Rest2} = compile_binding(Rest, S, <<>>),
 	Name = binary_to_atom(NameBin, utf8),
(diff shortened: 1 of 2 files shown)


fp:
The CI test results look to be the same as what I see on `master`, with my updated test passing in `http_SUITE`.  Is there anything I can/should do to resolve this?

essen:
Don't worry, if it works on your machine it's fine, I have to make circleci and others just test against a single release or they timeout. I'll have buildkite enabled on this project in the next week or two to test across everything.

fp:
@essen is there any way this could get merged to master?

essen:
I am minutes away from releasing 2.0.0-pre.4, and will take a look at existing tickets/PRs after that point. This PR looks ready so it shouldn't be too long after that (depends on my schedule).

Sorry for the long wait.

essen:
Think the test would be better added here: https://github.com/fp/cowboy/blob/be0aa758cd0451a0f12f40ea4d5aac74284fcbd2/src/cowboy_router.erl#L365

motobob:
@essen so if the tests are moved can this be a candidate for `2.0.0-pre10` or any next release?

essen:
It's not a priority (finalizing the interface is) but it's definitely something I will get to at some point, possibly after 2.0 is released though.

essen:
I have merged the PR mostly as-is and this will be in the next version. Thanks!

motobob:
thank you

PR #935, opened 2016-01-19
Add default_file parameter for cowboy_static

glejeune:
This allows you to specify a default file when using cowboy_static with
dir or priv_dir. Thus the path will be completed with the default file.

Example : 
```
{"/[...]", cowboy_static, {priv_dir, static_world, "", 
 [{mimetypes, cow_mimetypes, all}, {default_file, "index.html"}]}}
```
Thus http://localhost:8080/howdy/ will do the same as http://localhost:8080/howdy/index.html

### DIFF
--- a/examples/static_world/priv/howdy/index.html
+++ b/examples/static_world/priv/howdy/index.html
@@ -1,0 +1,1 @@
+<h1>Head 'em up, move 'em out.</h1>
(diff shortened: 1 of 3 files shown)


essen:
Out of scope of cowboy_static, sorry. You can easily do the same with a simple middleware between router and handler. If anything gets into Cowboy it would have to be in middleware form.

PR #938, opened 2016-01-24
Clarify naming in multipart example

rlipscombe:
In response to http://erlang.org/pipermail/erlang-questions/2016-January/087395.html, change the example slightly


essen:
Done locally with LastBodyChunk and BodyChunk instead. Thanks!

PR #941, opened 2016-02-02
improve handling empty binary performance in body_recv

bigwhite37:
We call the function "body_recv" when "buffer" in "http_req" is empty, so I think it shoud add a case for empty binary and it will get some optimization in our project which depends on cowboy.


essen:
How did you benchmark it, where is the benchmark code, where are the results? And why did you break the whitespace?

bigwhite37:
Breaking the whitespace is my missing.
I use `eprof:start/1` at the beginning of `cowboy_req:body/2` and `eprof:stop_profiling` written the end of it.
The function `cowboy_req:body_recv/3` will be called only when `buffer` in `http_req` is empty in `cowboy_req:body_loop/5`. However the code `<< Buffer/binary, Data/binary >>` in `cowboy_req:body_recv/5` will construct a new binary and the `Data` will be GC later which leads time wasting when `Buffer` is empty.

essen:
Sure but does it make any difference in a real world scenario and how much of a difference does it make?

bigwhite37:
A project, which used as the interface of storage system by CDN, uses cowboy to handle http requests. When I get a 51M's PUT, `cowboy_req:body/2` gets multi-call in a loop, use `<< Buffer/binary, Data/binary >>` getting 13000 - 37000 ms cost counting by `eprof`. Use my code, getting costs mostly below 5000 ms, but rarely getting same cost with nowadays code.

PR #945, opened 2016-02-07
Websocket compression

vans163:
Lets get this started..

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,2 +1,2 @@
-{deps, [{cowlib,".*",{git,"https://github.com/ninenines/cowlib","master"}},{ranch,".*",{git,"https://github.com/ninenines/ranch","1.1.0"}}]}.
+{deps, [{cowlib,".*",{git,"https://github.com/vans163/cowlib","websocket_compression"}},{ranch,".*",{git,"https://github.com/ninenines/ranch","1.1.0"}}]}.
 {erl_opts, [debug_info,warn_export_vars,warn_shadow_vars,warn_obsolete_guard,warn_export_all,warn_missing_spec,warn_untyped_record]}.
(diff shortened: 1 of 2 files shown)


vans163:
I am not fully understanding this client_context_takeover and sliding window size.  Does it put extra bytes on the wire?  Or is it purely handled server/client side?

Also what is the difference between a -15 and 15 sliding window?  I see that a 15bit sliding window will take around 120kb extra memory, but will significantly decrease after compression size vs a 8bit.

https://www.igvita.com/2013/11/27/configuring-and-optimizing-websocket-compression/

essen:
If I recall the thing about context is whether the context is kept for the whole connection or a new zlib context is created for each frames.

The plus/minus I don't recall off-hand, I probably left a comment in the code somewhere though.

essen:
Do run "make ct-ws" after making changes to confirm they work by the way. It requires autobahntestsuite and python to be installed.

vans163:
Dont see any related comment.  Let me try.

vans163:
Also do you know a the equivalent when using rebar to  make:all([load]). on the erl shell?

essen:
I don't use rebar.

If ct_run is not found then your Erlang installation is not correct and/or the paths are not set right.

vans163:
Cowboy has the rebar.config though? How do you then make changes to cowboy.erl files and hotload them?

Strange I installed the official package from erlang repo.

It seems setting window size to 15 crashes with no error (c driver crash?).  -15 works fine.

essen:
The rebar.config is there only for compatibility, it's generated from the Makefile.

I don't really do hot loading, I usually just run tests.

vans163:
The test suite took 30 minutes to run for me, is there a way to speed it up? Completed with no errors.

vans163:
Very interesting how the sliding window works with a msg that just says ping.

```
[No compression]
0000: 8104 7069 6e67  ..ping

[Compression -15 sliding]
0000: c106 2ac8 cc4b 0700  ..*..K..
0000: c105 2a00 6200 00 ..*.b..
0000: c104 0261 0000  ...a..
0000: c104 0261 0000  ...a..
```

essen:
Most of the test suite is spent in compression tests. If there's really a bottleneck in zlib, then that's probably why it takes so long. :-)

essen:
I've pushed configuration of Websocket deflate options. I will do the compression bit later. Closing since there's a ticket for it at ninenines/cowboy#1105 Thanks!

Issue #947, opened 2016-02-11
Make Cowboy CORS friendly


manifest:
Maybe we could make Cowboy a bit more CORS friendly?
By now, to enable CORS we should work with raw headers, copy-pasting code like this from one header to another:
```erlang
options(Req, State) ->
  Req1 = cowboy_req:set_resp_header(<<"access-control-max-age">>, <<"1728000">>, Req0),
  Req2 = cowboy_req:set_resp_header(<<"access-control-allow-methods">>, <<"HEAD, GET, POST">>, Req1),
  Req3 = cowboy_req:set_resp_header(<<"access-control-allow-headers">>, <<"content-type, authorization">>, Req2),
  Req4 = cowboy_req:set_resp_header(<<"access-control-allow-origin">>, <<$*>>, Req3),
  {ok, Req, State}.
```
In this way we do not actually care about allowed header's value types. For instance, "access-control-allow-methods" and "access-control-allow-headers" should only accept a list of binaries, while only a numeric value makes sense for the "access-control-max-age" header, "*" origin cannot be used for a resource that supports credentials. Ignoring this information leads to misspelling errors, breaking CORS responses and isn't flexible for working with.

My proposal is to add two new functions (`set_cors_headers/2` and `set_cors_preflight_headers/2`) to the `cowboy_req` module extending functionality of `set_resp_header/3`.

```erlang
-type cors_header()
  :: {origin, binary() | list(binary())}.

-type cors_preflight_header()
  :: {age, max | non_neg_integer()}
   | {methods, list(binary())}
   | {headers, list(binary())}
   | cors_header().

-spec set_cors_headers(list(cors_header()), Req) -> Req when Req :: cowboy_req:req().
-spec set_cors_preflight_headers(list(cors_preflight_header()), Req) -> Req when Req :: cowboy_req:req().
```

In my projects, I'm using this simple syntax sugar in REST handlers:

```erlang
allowed_methods(Req, State) ->
  Req2 = cowboy_req:set_cors_headers([{origin, <<$*>>}], Req),
  {[<<"HEAD">>, <<"GET">>, <<"POST">>, <<"OPTIONS">>], Req2, State}.

options(Req, State) ->
  Headers =
    [ {age, max},
      {origin, <<$*>>},
      {methods, [<<"HEAD">>, <<"GET">>, <<"POST">>]},
      {headers, [<<"content-type">>, <<"authorization">>]} ],
[...]

PR #949, opened 2016-02-15
Add support of CORS

manifest:
Pull request for issue [947](https://github.com/ninenines/cowboy/issues/947) "Make Cowboy CORS friendly".

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -67,6 +67,8 @@
 -export([has_resp_header/2]).
 -export([has_resp_body/1]).
 -export([delete_resp_header/2]).
+-export([set_cors_headers/2]).
+-export([set_cors_preflight_headers/2]).
 -export([reply/2]).
 -export([reply/3]).
 -export([reply/4]).
@@ -305,6 +307,8 @@ parse_header_fun(<<"accept">>) -> fun cow_http_hd:parse_accept/1;
 parse_header_fun(<<"accept-charset">>) -> fun cow_http_hd:parse_accept_charset/1;
 parse_header_fun(<<"accept-encoding">>) -> fun cow_http_hd:parse_accept_encoding/1;
 parse_header_fun(<<"accept-language">>) -> fun cow_http_hd:parse_accept_language/1;
+parse_header_fun(<<"access-control-request-headers">>) -> fun cow_http_hd:parse_access_control_request_headers/1;
+parse_header_fun(<<"access-control-request-method">>) -> fun cow_http_hd:parse_access_control_request_method/1;
 parse_header_fun(<<"authorization">>) -> fun cow_http_hd:parse_authorization/1;
 parse_header_fun(<<"connection">>) -> fun cow_http_hd:parse_connection/1;
 parse_header_fun(<<"content-length">>) -> fun cow_http_hd:parse_content_length/1;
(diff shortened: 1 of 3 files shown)


essen:
Thanks, a few comments. Note however that this will break in a week or two when I finally merge back my current set of changes (huge).

* I will not merge until I review the CORS specs, so please be patient
* I will not merge until there's tests or I have time to write some; those must be based on real-world use cases and added to a relevant Common Test suite (or a new suite must be created, probably this)
* It seems that there's parsing code in this PR. Header parsing code should be in cow_http_hd in cowlib (along with tests and benchmarks)
* I am not sure the function binary_join is necessary
* Use maps, the whole proplists converted to a record is completely unnecessary

manifest:
I will write the tests. I will also replace proplists on maps, you're right it's a better choice.
As for headers, I see that I can reuse `token_list` function from `cow_http_hd` module to parse comma separated lists in headers more effectively. But I didn't see something to build such a list (like this, "Access-Control-Allow-Methods: PUT, DELETE, XMODIFY"), so I could throw `binary_join` away.

essen:
Hey, I will try to get to this this week-end. I finally read the CORS spec so we can move forward.

This seems like a good first step, but I think we should also better integrate CORS support with cowboy_rest. Most of the info can be filled automatically there, except the origin check, max-age and exposed-headers, so it could be very nice to have.

src/cowboy_req.erl
```diff
@@ -666,6 +671,126 @@ delete_resp_header(Name, Req=#http_req{resp_headers=RespHeaders}) ->
...
+-spec match_cors_headers([binary()], [binary()]) -> [binary()].
+match_cors_headers(L, AllowedHeaders) ->
+	[case lists:member(Header, AllowedHeaders) of
+		false -> throw({nomatch_header, Header, AllowedHeaders});
+		_ -> Header
+	end || Header <- L].
+
+-spec match_cors_credentials(boolean(), {binary(), binary(), 0..65535} | reference() | '*') -> boolean().
+match_cors_credentials(true, '*') ->
+	throw({bad_credentials, true, '*'});
```
essen: Guessing this is for 'The string "*" cannot be used for a resource that supports credentials.' but I'm not sure this is very useful, more useful would be sending Origin back instead of "*" if credentials are allowed.

manifest: Actually, it's how it works now. User will get an Origin back if credentials are allowed. The only way he can get the "*" in the response the parser didn't fully recognize the Origin value or the value was "null" string, thus the parser returned an reference that can't be returned to the user. So we need this validation for the last case.

hachreak:
Hi at all,
are there news about CORS for cowboy?
Thanks :)

manifest:
@hachreak, the main conversation is [here](https://github.com/ninenines/cowboy/issues/947).

hachreak:
hi @manifest, I saw that the PR for `cowlib` have been merged.
It's mean that also this PR will be merged too? 
Thanks :)

manifest:
@hachreak, this PR is currently obsoleted. You can implement the support of CORS in a way described in the last few comments [here](https://github.com/ninenines/cowboy/issues/947), using new functions of cowlib. Will it work for you?

hachreak:
if you have time, can you add a example that show how it works? :)
This will be really useful to anyone want use it.

manifest:
@hachreak, have you seen an [example](https://github.com/ninenines/cowboy/issues/947) here? Is it not clear how you can implement CORS support from that example or you mean an example in the docs?

hachreak:
I mean an example inside [examples/](https://github.com/ninenines/cowboy/tree/master/examples) directory.

manifest:
Ok, will do it.

hachreak:
@manifest thanks a lot! :D

PR #953, opened 2016-02-27
fix bug: change + to '%2B'

tianjipeng:
fix issue#942: cowboy_static won't serve file with + in the path


essen:
This is definitely not the right fix. The protocol parser shouldn't do any substitution, it's for the path parsing code to do. A proper fix starts with adding the required function to Cowlib, and then using it where appropriate.

tianjipeng:
Ok, I will try what you said.

PR #991, opened 2016-05-25
gitignore: add examples/ dependency files

katsel:
I built some of the examples and found files missing from `.gitignore`.


essen:
Thanks!

PR #992, opened 2016-05-29
cowboy:start_http does not work and gives a run-time error.

MatthewKantemir:
When I made the following change code runs without any problem.


essen:
Yes that example has not been converted yet.

essen:
The example has been converted and a test has been added. Thanks!

PR #997, opened 2016-06-22
don't crash on bad cookie header

sstrigler:
For a cookie header like "Cookie: foo" cowboy would just crash when calling cowboy_req:cookie/3.

```
17:56:54.312 [error] Ranch listener needle_http_listener terminated in needle_notes_h:forbidden/2 with reason: no case clause matching {error,badarg} in cowboy_req:cookie/3 line 429
17:56:54.314 [error] Error in process <0.1211.0> on node needle@eli with exit value:
{[{reason,{case_clause,{error,badarg}}},{mfa,{needle_notes_h,forbidden,2}},{stacktrace,[{cowboy_req,cookie,3,[{file,"_build/default/lib/cowboy/src/cowboy_req.erl"},{line,429}]},{needle_rest,check_authenticated,3,[{file,"/Users/zeank/src/JPCatholic/needle/_build/dev/lib/needle/src/needle_rest.erl"},{line,36}]},{cowboy_rest,call,3,[{file,"/Users/zeank/src/JPCatholic/needle/_build/default/lib/cowboy/src/cowboy_rest.erl"},{line,976}]},{cowboy_rest,expect,6,[{file,"/Users/zeank/src/JPCatholic/needle/_build/default/lib/cowboy/src/cowboy_rest.erl"},{line,960}]},{cowboy_protocol,execute,4,[{file,"/Users/zeank/src/JPCatholic/needle/_build/default/lib/cowboy/src/cowboy_protocol.erl"},{line,442}]}]},{req,[{socket,#Port<0.119171>},{transport,ranch_tcp},{connection,keepalive},{pid,<0.1211.0>},{method,<<"GET">>},{version,'HTTP/1.1'},{peer,{{127,0,0,1},49319}},{host,<<"localhost">>},{host_info,undefined},{port,8080},{path,<<"/classes/123/sessions/1232/notes">>},{path_info,undefined},{qs,<<>>},{qs_vals,undefined},{bindings,[{session,<<"1232">>},{class,<<"123">>}]},{headers,[{<<"host">>,<<"localhost:8080">>},{<<"user-agent">>,<<"curl/7.43.0">>},{<<"accept">>,<<"*/*">>},{<<"cookie">>,<<"foo">>}]},{p_headers,[]},{cookies,undefined},{meta,[]},{body_state,waiting},{buffer,<<>>},{multipart,undefined},{resp_compress,false},{resp_state,waiting},{resp_headers,[{<<"access-control-allow-credentials">>,<<"true">>},{<<"access-control-allow-origin">>,<<"*">>}]},{resp_body,<<>>},{onresponse,undefined}]},{state,#{}}],[{cowboy_rest,expect,6,[{file,"/Users/zeank/src/JPCatholic/needle/_build/default/lib/cowboy/src/cowboy_rest.erl"},{line,960}]},{cowboy_protocol,execute,4,[{file,"/Users/zeank/src/JPCatholic/needle/_build/default/lib/cowboy/src/cowboy_protocol.erl"},{line,442}]}]}
```

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -427,6 +427,8 @@ cookie(Name, Req) when is_binary(Name) ->
 	-> {binary() | Default, Req} when Req::req(), Default::any().
 cookie(Name, Req=#http_req{cookies=undefined}, Default) when is_binary(Name) ->
 	case parse_header(<<"cookie">>, Req) of
+                {error, badarg} ->
+                        {Default, Req#http_req{cookies=[]}};
 		{ok, undefined, Req2} ->
 			{Default, Req2#http_req{cookies=[]}};
 		{ok, Cookies, Req2} ->


essen:
Thanks.

essen:
Wondering what would send "Cookie: foo" and if we really want to silently ignore any bad cookie value?

I think it would make more sense to crash and let the developer catch exceptions if needed, rather than silently drop it, unless a popular software is generating these cookie strings (like we do for empty cookie values).

sstrigler:
My initial concern was that it is hard to debug and understand what's going on when all you've got is that crash in your logs. That someone might have sent a bad cookie header. In my case I knew why because I was doing the request manually. But then I thought of a system in production.

sstrigler:
That said, maybe just having that reported issue around is already helpful enough?

essen:
In Cowboy 2 this function is renamed into "parse_cookie" and its name and docs will imply that parsing is done and an exception will be thrown on error. But for 1.0 it's indeed confusing. Let's keep it open for now and see what happens. Thanks!

sstrigler:
Ah, nice! :)

essen:
Won't merge for 1.1, too much uncertainty, I need to investigate more thoroughly than time allows. Thanks!

essen:
Cowboy 1.x is old. Closing, thanks!

PR #998, opened 2016-06-25
send resp_headers with websocket responses

camshaft:
This fix is needed to support subprotocol negotiation (with the `Sec-WebSocket-Protocol` header) and sending cookies (with the `Cookie` header) in websockets as outlined in [the spec](https://whatwg.org/specs/web-socket-protocol/), section 1.3. It looks like it was supported in Cowboy v1, according to #672.


essen:
Thanks! Looks like it got removed by mistake indeed. Please add a test to ws_SUITE so it doesn't happen again.

camshaft:
Alright. Let me know if that looks ok.

essen:
Thanks, will try to remember this PR tomorrow. :-)

essen:
Merged with tweaks, thanks!

PR #999, opened 2016-06-30
add license title

waldyrious:
It's not strictly required, but it's useful metadata, and part of the recommended license template text (see http://choosealicense.com/licenses/isc/ and https://opensource.org/licenses/isc-license)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+ISC License
+
 Copyright (c) 2011-2014, Loïc Hoguin <essen@ninenines.eu>
 
 Permission to use, copy, modify, and/or distribute this software for any


essen:
I will not merge this. I'd rather people read the three lines instead of assuming they know what it's about.

On the other hand this made me notice the license information is missing from the documentation. That will be fixed on the next push in a few days.

Thanks.

waldyrious:
Sure, I understand your point, and adding a title is of course entirely optional. May I add though that reading the license and understanding it is not enough: as most people lack the adequate legal background, they have to rely on trustworthy organizations to validate specific licenses (OSI, FSF) as compatible and not including (or lacking) small details they may overlook otherwise. And for this end, the only way to ensure a license is compliant is comparing it word by word with a few samples, which may be tedious if they don't already have a clear idea of what license (or even kind of license, e.g. permissive vs. copyleft) they must compare the text to. By providing a title you not only make this task hugely easier for humans, but also make this repo more discoverable and machine-processable by tools that can't afford to do full-text matching across thousands of repositories vs. dozens of possible licenses.

Given the above, I'd strongly urge you to reconsider. But again, it's ultimately your decision. Thanks anyway for choosing one of the simplest and most readable licenses around :)

essen:
The title will be mentioned in the documentation.

PR #1001, opened 2016-07-02
Add CORS example

manifest:
An example application for [this](https://github.com/ninenines/cowboy/issues/947) issue.


essen:
Good idea, thanks. Sorry it takes so long, I'm forcing my way through 2.0.0-pre.4 and will look at existing tickets/PRs once my set of changes is in, since I'm still breaking things.

PR #1006, opened 2016-07-18
strip whitespace at end of header field-value

glassbead0:
- requests with ows after the port in the field-value
  fail with a 400.
- this fix is to comply with
  https://tools.ietf.org/html/rfc7230#section-3.2.4


essen:
What client requires this?

alakra:
@essen This [device](https://www.universal-devices.com/residential/isy994i-series/) actual behaves this way when it makes HTTP requests.

essen:
Thanks.

essen:
For the record this is only an issue in 1.x, not in master.

essen:
This is missing a test so I won't include in 1.1. I'll reschedule it for master. Thanks!

essen:
Oh my bad, this isn't an issue in master. Closing then, sorry!

PR #1007, opened 2016-07-25
Fix if-modified-since in rest handlers (without last_modified callback implemented)

Vorticity-Flux:
In cowboy_rest.erl function if_modified_since/3 expects last_modified/2 to return either {no_call, Req, State} or {LastModified, Req2, State2}. If last_modified/2 callback is not defined in handler 
last_modified/2 function in cowboy_rest.erl returns {undefined, Req, State} and not  {no_call, Req, State}.
As undefined atom is always smaller then any timestamp not_modified/2 is always called.

Test case:
1. Rest handler without last_modified callback implemented.
2. Request with if-modified-since header and valid date in the past.

Current result: 304 Not modified reply
Expected: Full reply as resource could have been modified.

### DIFF
--- a/src/cowboy_rest.erl
+++ b/src/cowboy_rest.erl
@@ -626,7 +626,7 @@ if_modified_since_now(Req, State, IfModifiedSince) ->
 
 if_modified_since(Req, State, IfModifiedSince) ->
 	try last_modified(Req, State) of
-		{no_call, Req2, State2} ->
+		{undefined, Req2, State2} ->
 			method(Req2, State2);
 		{LastModified, Req2, State2} ->
 			case LastModified > IfModifiedSince of


christopheradams:
I'd like to see a test added that surfaces this bug.

christopheradams:
I tried this out and it looks correct to me. `last_modified` will return either `{undefined, Req, State}` or `{LastModified, Req, State}`. I don't see it ever returning `no_call`.

However, my guess is that it would be better to rebase this PR on `master`.

essen:
I still need a test. :-) Master is better but I can do with either.

christopheradams:
Not the OP, but I'd be willing to write a test for this, given a little guidance. Is there a todo in the rfc test where this would fit in? Another approach would be to create a `rest_conditional_requests` project in examples that could be used to drive the examples SUITE for this test.

essen:
At the moment the http_SUITE is the best place for REST tests. I'm not quite that far in the RFC tests right now.

essen:
Merged the PR #1030 which had this fix plus the test added. Thanks!

PR #1008, opened 2016-07-26
fix echo_get example

sergey-ivlev:
fix undefined case


essen:
I added a test and merged, thanks!

PR #1022, opened 2016-08-31
Update websocket documentation: based on https://github.com/ninenines…

alexprut:
Update websocket documentation based on the last API https://github.com/ninenines/cowboy/blob/master/examples/websocket/src/ws_handler.erl


essen:
Merged, thanks!

PR #1030, opened 2016-10-06
Fix if_modified_since match of last_modified

christopheradams:
Correct expected return type from `no_call` to `undefined` in
if_modified_since when last_modified callback is not defined. Add an
http_SUITE test to catch regressions.

Resources with undefined `last_modified` should return "200" rather than "304" when the request includes a "If-Unmodified-Since" header.

This implements the same fix as PR  #1007 but rebased on master and with a corresponding test that fails if the fix isn't there.

### DIFF
--- a/src/cowboy_rest.erl
+++ b/src/cowboy_rest.erl
@@ -782,7 +782,7 @@ if_modified_since_now(Req, State, IfModifiedSince) ->
 
 if_modified_since(Req, State, IfModifiedSince) ->
 	try last_modified(Req, State) of
-		{no_call, Req2, State2} ->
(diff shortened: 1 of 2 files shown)


essen:
Looks good.

christopheradams:
JSON was addressed but has anyone looked at XHTML? Does anyone use it for output?

essen:
Keeping you waiting, Cowboy is next on my list (updating CI configs, adding mirrors and so on for all projects in order).

essen:
Got this running in CI. Probably won't inform you about status since I started it manually.

essen:
Merged, thanks!

PR #1050, opened 2016-11-03
Fix http2 terminate error when socket is closed before fetching the peername in init/8.

mworrell:
As the description.

This happens when using self signed certs.
The client closes the connection if the cert is not yet approved (using an exception) by the user.


essen:
Thanks. Test case would be good (can use a custom protocol module that waits for the socket to be closed before continuing).

mworrell:
A whole end-to-end test or just a simple test case where we supply an invalid (closed/whatever) port to trigger the error during the fetch of the peer name?

PR #1053, opened 2016-11-13
Add functions for getting and setting response headers

johlo:
Useful when writing middleware.

### DIFF
--- a/doc/src/manual/cowboy_req.asciidoc
+++ b/doc/src/manual/cowboy_req.asciidoc
@@ -67,7 +67,10 @@ Request body:
 Response:
 
 * link:man:cowboy_req:set_resp_cookie(3)[cowboy_req:set_resp_cookie(3)] - Set a cookie
+* link:man:cowboy_req:resp_header(3)[cowboy_req:resp_header(3)] - Access the named HTTP header set for the response
+* link:man:cowboy_req:resp_headers(3)[cowboy_req:resp_headers(3)] - Access HTTP headers set for the response
 * link:man:cowboy_req:set_resp_header(3)[cowboy_req:set_resp_header(3)] - Set a response header
+* link:man:cowboy_req:set_resp_headers(3)[cowboy_req:set_resp_headers(3)] - Set several response headers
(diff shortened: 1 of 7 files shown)


essen:
Thanks, but it's missing tests. Check test/req_SUITE.erl

johlo:
@essen amended tests to the commit.

essen:
Tests are running! Thanks.

johlo:
Anything else needs fixing for merge?

essen:
I just need time to look into it.

essen:
I notice that the docs are missing. Did you write them and forget to git add? If not, it shouldn't be difficult to write the two man pages based on the other cowboy_req functions that have them ready.

essen:
Cheers, just in time, tomorrow is open source day.

johlo:
Sorry, forgot to write them, amended now. 
Note that I could build the docs so they might be a bit off. The docbookx DTD isn't fetched when doing `make docs`

essen:
Rebase against master and try `make asciidoc-manual`. It doesn't use DocBook for the man pages anymore.

johlo:
with `asciidoc-manual` it produced the man pages (although it fails with unspecified error in later stage). 
Let me know if you want me to change the wording.

essen:
Error seems to be in cowboy_rest, no worries.

essen:
Can you give me a use case for getting all response headers? I can get why a `resp_header` function would be useful, but am having problems with a `resp_headers` function.

Also they can't be obtained through pattern matching, that field is undocumented.

johlo:
It gives you a way of inspecting which response headers were set without having to know all the possible headers that could be set. Also I'd expect it since the bindings/headers exist.

johlo:
I removed the pattern matching suggestion from the doc.

essen:
Sure. Any chance you could add a resp_header at the same time then? It'll make more sense as a whole.

johlo:
Amended with resp_header/{2,3}, with test cases and manual.

essen:
Thanks. I should be able to get to it this week-end, documenting response functions now.

essen:
Merging this now, great PR thanks.

essen:
Merged with edits in the next two commits. Thanks!

PR #1059, opened 2016-12-10
partial implementation of rfc 7233 for cowboy_rest

s-kostyaev:
For full imlementation multipart/byteranges reply need to be implemented in cowboy.

### DIFF
--- a/src/cowboy_rest.erl
+++ b/src/cowboy_rest.erl
@@ -18,6 +18,8 @@
 -behaviour(cowboy_sub_protocol).
 
 -export([upgrade/6]).
+-export([range_satisfiable/2,
+         range_satisfiable/3]).
 
 %% Common handler callbacks.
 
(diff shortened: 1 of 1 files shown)


s-kostyaev:
You can see how to use it in this example:
```erlang
%% callback for content type with ranges support
get_file(Req, State = #state{filepath = FilePath, is_range = true, range_offset = Offset, range_bytes = Bytes}) ->
    SendFile = fun(Socket, Transport) ->
        case Transport:sendfile(Socket, FilePath, Offset, Bytes) of
            {ok, _} -> ok;
            {error, closed} -> ok;
            {error, etimedout} -> ok
        end
               end,
    {{stream, Bytes, SendFile}, Req, State};
get_file(Req, State = #state{filepath = FilePath, file_info = #file_info{size = Size}}) ->
    SendFile = fun(Socket, Transport) ->
        case Transport:sendfile(Socket, FilePath) of
            {ok, _} -> ok;
            {error, closed} -> ok;
            {error, etimedout} -> ok
        end
               end,
    {{stream, Size, SendFile}, Req, State}.

%% callback for using http ranges
range_satisfiable(Req, State = #state{file_info = #file_info{size = Size}}) ->
    case cowboy_rest:range_satisfiable(Req, Size) of
        {true, Req1, Offset, Bytes} ->
            {true, Req1, State#state{is_range = true, range_offset = Offset,
                                     range_bytes = Bytes}};
        _ ->
            {false, Req, State#state{is_range = false}}
    end.
```

essen:
Multipart is already supported so I'm not sure what you mean.

s-kostyaev:
@essen please add link to doc. I need reply from cowboy with multipart, not receive multipart in cowboy.

essen:
cow_multipart, no docs yet.

s-kostyaev:
Thanks. We use branch 1.0. Maybe we complete rfc 7233 implementation later. Or you can merge this partial implementation and complete it yourself.

s-kostyaev:
For now this implementation works fine with `<audio>` tag in firefox and google chrome.

essen:
I've started working on a full implementation of RFC7233. I may or may not take inspiration or some code from this PR. In the meantime I'll close it. Thanks!

PR #1064, opened 2016-12-16
Return status 431 if the request header field is too large

### LINKED ISSUE #1063: Return status 431 if the request header fields are too large (RFC6585)
RFC6585 says:

> 431 Request Header Fields Too Large (RFC6585 5, RFC6585 7.3)
> The request's header fields are too large.
> When rejecting a single header, the response should detail which header was at fault.
> When an attack is detected it is recommended to drop the connection directly instead of sending this response.

However, when a header name or value are too large, Cowboy simply returns 400. Returning a more appropriate status code should also help developers who ran into this limit by accident, for example because they end-up creating a large cookie by mistake.

I will be glad to send a PR if agreed this would be an improvement.

josevalim:
This commit changes Cowboy to follow RFC6585.

Closes #1063

### DIFF
--- a/test/http_SUITE.erl
+++ b/test/http_SUITE.erl
@@ -307,6 +307,20 @@ echo_body(Config) ->
 	end || Size <- lists:seq(MTU - 500, MTU)],
 	ok.
 
+%% Check if sending request whose header name is bigger than 64 bytes causes 431
+echo_body_max_header_name_length(Config) ->
+	ConnPid = gun_open(Config),
+	Ref = gun:post(ConnPid, "/echo/body", [{binary:copy(<<$a>>, 32768), <<"bad">>}], << "echo=name" >>),
+	{response, fin, 431, _} = gun:await(ConnPid, Ref),
+	ok.
+
+%% Check if sending request whose header name is bigger than 64 bytes causes 431
+echo_body_max_header_value_length(Config) ->
+	ConnPid = gun_open(Config),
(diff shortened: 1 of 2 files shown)


test/http_SUITE.erl
```diff
@@ -307,6 +307,20 @@ echo_body(Config) ->
 	end || Size <- lists:seq(MTU - 500, MTU)],
 	ok.
 
+%% Check if sending request whose header name is bigger than 64 bytes causes 431
+echo_body_max_header_name_length(Config) ->
+	ConnPid = gun_open(Config),
+	Ref = gun:post(ConnPid, "/echo/body", [{binary:copy(<<$a>>, 32768), <<"bad">>}], << "echo=name" >>),
```
josevalim: @essen I had to send a much larger header name because a binary with size right after the limit (which would be 65 bytes) was not triggering the error. Or I am missing something obvious or there is some buffer handling that is causing it to not check the whole header name.

The same applies to the header value below.

essen: Most limits in Cowboy are soft limits. If we have already read the data from the socket and this data happened to be over the limit, we don't throw an error. This limit is about not reading endlessly from the socket.

essen:
Launching tests, they will fail for unrelated reasons, and I will look more closely next week. Cheers!

essen:
I have merged this and fixed the test in 42c95b1584c245d73e6ddc052a922c8e7c68dee2. Closing, thanks!

PR #1066, opened 2016-12-21
Document correction

sumiyoshi:
Is this incorrect?

### DIFF
--- a/doc/src/guide/handlers.asciidoc
+++ b/doc/src/guide/handlers.asciidoc
@@ -29,9 +29,9 @@ We need to use the Req object to reply.
 [source,erlang]
 ----
 init(Req0, State) ->
-    Req = cowboy_req:reply(200, [
+    Req = cowboy_req:reply(200, #{
         {<<"content-type">>, <<"text/plain">>}
-    ], <<"Hello World!">>, Req0),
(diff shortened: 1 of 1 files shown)


essen:
Yep it hasn't been updated yet.

sumiyoshi:
Thank you for reply

doc/src/guide/handlers.asciidoc
```diff
@@ -29,9 +29,9 @@ We need to use the Req object to reply.
 [source,erlang]
 ----
 init(Req0, State) ->
-    Req = cowboy_req:reply(200, [
+    Req = cowboy_req:reply(200, #{
         {<<"content-type">>, <<"text/plain">>}
```
essen: Yep!

sumiyoshi: Is not ->, =>?

essen:
Still not quite right, but don't worry I will take care of it. Almost finished updating the cowboy_req manual.

essen:
Do leave open so I don't forget. :-) Cheers.

sumiyoshi:
OK！

essen:
This is done on my local copy and will be pushed later today. Thanks!

PR #1067, opened 2016-12-22
Updates cowboy_http documents changelog

shezarkhani:
The changelog had a wrong reference to an option that was updated.
`max_header_request_line_length` -> `max_request_line_length`


essen:
Merged, thanks!

PR #1068, opened 2016-12-25
support configuring http2 receive timeout

tony612:


essen:
This is missing a test at the very least, updating the documentation, and should not introduce a new function when we can do the maps:get in the function directly.

essen:
Closing in favor of #1141.

PR #1070, opened 2016-12-29
cowboy_static: Add support for files in EZ archives

dumbbell:
This allows to serve content from eg. the `priv` directory of a compressed Erlang application.


dumbbell:
I prepared a patch for the 1.0.x branch as well and tested it successfully with the RabbitMQ management plugin.

```diff
diff --git a/src/cowboy_static.erl b/src/cowboy_static.erl
index fae4568..f0f7e2a 100644
--- a/src/cowboy_static.erl
+++ b/src/cowboy_static.erl
@@ -37,7 +37,7 @@
 
 -include_lib("kernel/include/file.hrl").
 
--type state() :: {binary(), {ok, #file_info{}} | {error, atom()}, extra()}.
+-type state() :: {binary(), {ok, on_disk | in_ez, #file_info{}} | {error, atom()}, extra()}.
 
 -spec init(_, _, _) -> {upgrade, protocol, cowboy_rest}.
 init(_, _, _) ->
@@ -111,7 +111,20 @@ fullpath([Segment|Tail], Acc) ->
 	fullpath(Tail, [Segment|Acc]).
 
 rest_init_info(Req, Path, Extra) ->
-	Info = file:read_file_info(Path, [{time, universal}]),
+	Info = case file:read_file_info(Path, [{time, universal}]) of
+		{ok, I} ->
+			{ok, on_disk, I};
+		{error, enotdir} ->
+			%% The file might be inside an Erlang
+			%% application archive (.ez).
+			PathS = binary_to_list(Path),
+			case erl_prim_loader:read_file_info(PathS) of
+				{ok, I} -> {ok, in_ez, I};
+				error   -> {error, enotdir}
+			end;
+		Other ->
+			Other
+	end,
 	{ok, Req, {Path, Info, Extra}}.
 
 -ifdef(TEST).
@@ -212,11 +225,11 @@ malformed_request(Req, State) ->
 -spec forbidden(Req, State)
 	-> {boolean(), Req, State}
 	when State::state().
-forbidden(Req, State={_, {ok, #file_info{type=directory}}, _}) ->
+forbidden(Req, State={_, {ok, _, #file_info{type=directory}}, _}) ->
 	{true, Req, State};
 forbidden(Req, State={_, {error, eacces}, _}) ->
 	{true, Req, State};
-forbidden(Req, State={_, {ok, #file_info{access=Access}}, _})
+forbidden(Req, State={_, {ok, _, #file_info{access=Access}}, _})
 		when Access =:= write; Access =:= none ->
 	{true, Req, State};
 forbidden(Req, State) ->
@@ -242,7 +255,7 @@ content_types_provided(Req, State={Path, _, Extra}) ->
 -spec resource_exists(Req, State)
 	-> {boolean(), Req, State}
 	when State::state().
-resource_exists(Req, State={_, {ok, #file_info{type=regular}}, _}) ->
+resource_exists(Req, State={_, {ok, _, #file_info{type=regular}}, _}) ->
 	{true, Req, State};
 resource_exists(Req, State) ->
 	{false, Req, State}.
@@ -252,7 +265,7 @@ resource_exists(Req, State) ->
 -spec generate_etag(Req, State)
 	-> {{strong | weak, binary()}, Req, State}
 	when State::state().
-generate_etag(Req, State={Path, {ok, #file_info{size=Size, mtime=Mtime}},
+generate_etag(Req, State={Path, {ok, _, #file_info{size=Size, mtime=Mtime}},
 		Extra}) ->
 	case lists:keyfind(etag, 1, Extra) of
 		false ->
@@ -271,7 +284,7 @@ generate_default_etag(Size, Mtime) ->
 -spec last_modified(Req, State)
 	-> {calendar:datetime(), Req, State}
 	when State::state().
-last_modified(Req, State={_, {ok, #file_info{mtime=Modified}}, _}) ->
+last_modified(Req, State={_, {ok, _, #file_info{mtime=Modified}}, _}) ->
 	{Modified, Req, State}.
 
 %% Stream the file.
@@ -280,7 +293,7 @@ last_modified(Req, State={_, {ok, #file_info{mtime=Modified}}, _}) ->
 -spec get_file(Req, State)
 	-> {{stream, non_neg_integer(), fun()}, Req, State}
 	when State::state().
-get_file(Req, State={Path, {ok, #file_info{size=Size}}, _}) ->
+get_file(Req, State={Path, {ok, on_disk, #file_info{size=Size}}, _}) ->
 	Sendfile = fun (Socket, Transport) ->
 		case Transport:sendfile(Socket, Path) of
 			{ok, _} -> ok;
@@ -288,4 +301,13 @@ get_file(Req, State={Path, {ok, #file_info{size=Size}}, _}) ->
 			{error, etimedout} -> ok
 		end
 	end,
+	{{stream, Size, Sendfile}, Req, State};
+get_file(Req, State={Path, {ok, in_ez, #file_info{size=Size}}, _}) ->
+	Sendfile = fun (Socket, Transport) ->
+		%% A static file in an Erlang
+		%% application archive (.ez).
+		PathS = binary_to_list(Path),
+		{ok, Bin, _} = erl_prim_loader:get_file(PathS),
+		Transport:send(Socket, Bin)
+	end,
 	{{stream, Size, Sendfile}, Req, State}.
```

gerhard:
How can I help with this PR? I don't have access to the checks that failed, that seemed like a good place to start.

essen:
There's no test for the feature being added so there's no test failing because of it.

dumbbell:
I didn't add any test for now because I would like an initial review on the approach. When we are ok on the feature, then I can work on a testcase.

dumbbell:
What about adding this support only when `cowboy_static` is called with `{priv_file, ...}` or `{priv_dir, ...}`?

essen:
Is it always a .ez file = an OTP application? In other words do you need something more than an explicit `{ez_file` and `{ez_dir` would provide?

dumbbell:
Using an explicit `ez_file` instead of `priv_file` supporting both ways means that the caller needs to also check how the application is made available.

As for the ".ez file == an OTP application", I don't know.

essen:
I need to look at this closer then. I'll be around Thursday.

dumbbell:
The [documentation on Erlang archives](http://erlang.org/doc/man/code.html#id108042) states that an archive can be an entire application or a part of an application (eg. you have files in `myapp-1.0/ebin` and files in `myapp-1.0.ez/myapp-1.0/priv`).

In the case of an escript, it can contain several applications and the escript name can ba anything. But in the "normal" case, the archive name must match the top-level directory name inside the archive, and the content must follow the standard layout of an application (so `myapp-1.0.ez` contains `myapp-1.0/{ebin,include,priv}` for instance).

When an application is available as a directory and an archive, second-levels directories (`ebin`, `priv`, etc.) are taken first in the directory then the archive in case of a conflict. The documentation doesn't tell what the code does when there is eg. conflicting `ebin` directories containing non-conflicting files. This is probably asking for troubles anyway :-)

I believe we can ignore the special case of escript here. For normal applications, I would expect that it is transparent to users of Cowboy. So if the caller says it wants to serve files using `priv_file` or `priv_dir` (eg. `{priv_file, myapp, <<"index.html">>, Extra}`).

If `file` and `dir` don't support serving files from an EZ archive, it's ok.

essen:
OK I went to check the documentation and the code to confirm a few things. I have a better idea of how this all works now. I have no problem with the general solution in this PR. There is however one issue and some minor changes to make it less intrusive.

The issue is that `erl_prim_loader:read_file_info` reads the MSDOS last modification time from the zip. This is encoded over 4 bytes (2 date 2 time), and does not contain timezone information. Further, the `zip:create` function will by default use the *local time* of the files it includes. This means that, as far as I can tell, the modification time we use fort he last-modified header will not be correct.

I am not sure what the correct solution is for this. I would be OK with *not* sending a last-modified in that case, at least for the time being. The last-modified header is deprecated anyway, and is ignored when an etag header is sent, so this will only affect old and badly written clients.

Other than that, the minor changes I mention would involve using a different tuple (maybe use `file` or `archive` replacing the `ok` in the tuple, so that we can match on that only in the relevant functions. We don't actually need to match on `ok` since we match on `#file_info` everywhere anyway.

dumbbell:
I was uncomfortable with the stats for files in the .ez archive. The other idea I had (but not implemented) was to use the modification date of the archive itself, instead of its content (and thus ignore the contained timestamps). What do you think?

essen:
And for the record, yes, limit to `priv_dir` and `priv_file`. The other options are meant to get files outside the release/applications.

dumbbell:
Hi!

I just pushed a new version of the patch, including testcases. Those tests rely on a provided application archive (in `static_handle_SUITE_data`).

The whole testsuite returns a lot of failures for me, perhaps I don't run it correctly? I simply use `make tests` without any environment preparation. At least, the patch show no regressions in existing testcases.

I have a patch for `examples_SUITE` (to not hardcode the `make` executable name) that I'll submit separately.

essen:
Cowboy has a lot of test failures currently. Work in progress! On the other hand `make ct-static_handler` should run without any failure.

dumbbell:
It does run successfully for me.

essen:
Merged, thanks!

PR #1072, opened 2017-01-03
Example for allow to upload file with any size

ethrbh:
hello Essen,

Here is the new PR for "allow to upload file with any size". Please check it out.

thanks,
/Robi


essen:
Hey, thanks for the PR. While it allows us to accept a file of any size, it is limited by the amount of memory available. I would prefer not to add an example like this. A better example would save to the file continuously and make sure the file is fully written and otherwise delete it. Closing, thanks!

PR #1076, opened 2017-01-17
examples_SUITE: Query system to find GNU Make executable

dumbbell:
... instead of hard-coding "make".

First, we check the value of `$MAKE`. If it's unset, we look for `gmake` in the `$PATH`. If it's missing, we assume it's `make`.

This fixes the testsuite where GNU Make is installed as `gmake`.


essen:
Merged, thanks!

PR #1080, opened 2017-01-25
cowboy_static: Add support for files in EZ archives (backport to 1.1x)

dumbbell:
If cowboy_static is initialized with `{priv_file, ...}` or `{priv_dir, ...}`, it is now able to read files from Erlang application .ez archives.

When serving a file from an archive, the #file_info{} comes from the archive, not the contained file, except for the size and type. The erl_prim_loader module is used to read the latter's #file_info{} and the actual file content (ie. sendfile(2) is not used in this case).

(cherry picked from commit 2166733628bbab0eb77eeed58bdf204727c48ec6)


dumbbell:
FTR, I tested this modified Cowboy successfully with RabbitMQ management plugin not being expanded (after modifying it to use `priv_file` and `priv_dir`).

essen:
Merged, will be pushed in a few. Just checking if there's anything interesting for 1.1.x right now.

PR #1089, opened 2017-02-01
Update ranch to 1.3.1 in rebar.config

### LINKED ISSUE #1088: Release 1.1.0 has inconsistent dependency on ranch
The recent [1.1.0](https://github.com/ninenines/cowboy/tree/1.1.0) has an inconsistent version pin for ranch. the Makefile pins it at [1.3.1](https://github.com/ninenines/cowboy/blob/1.1.0/Makefile#L18), but it's still set at [1.2.0](https://github.com/ninenines/cowboy/blob/1.1.0/rebar.config#L1) in the rebar.config.

My company uses rebar to package our Erlang HTTP services, and we were just looking at updating from 1.0.4 to 1.1.0, but we've found that ranch is still pulled down from `1.2.0`. This seems like an oversight. Is it possible to get a 1.1.1 tag that corrects this inconsistency?

nalundgaard:
Update to match the Makefile change in 3a240eb49b6322ec88985aa21014ca84a81157bb. Fixes #1088.


essen:
Thanks, merging/releasing tomorrow!

essen:
Done, thanks!

PR #1131, opened 2017-02-21
Add support for HTTP 451 (RFC 7725)

nalundgaard:
I noticed that this code seems to be missing from cowboy.

See [rfc7725 section 3](https://tools.ietf.org/html/rfc7725#section-3).

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -1316,6 +1316,7 @@ status(426) -> <<"426 Upgrade Required">>;
 status(428) -> <<"428 Precondition Required">>;
 status(429) -> <<"429 Too Many Requests">>;
 status(431) -> <<"431 Request Header Fields Too Large">>;
+status(451) -> <<"451 Unavailable For Legal Reasons">>;
 status(500) -> <<"500 Internal Server Error">>;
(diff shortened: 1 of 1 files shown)


nalundgaard:
@essen Any chance you guys could glance this? It seems pretty straightforward. If there's anything I am missing, I would be happy to update.

essen:
It's not a priority right now. The workaround is to send the binary (code + description) directly.

nalundgaard:
@essen For us, we have an internal framework built on top of cowboy (recently migrated from being on top of mochiweb) for our own integrated AAA. That thing has an implicitly integer return status that would be a huge pain to unthread. For our part, we've pinned our application against my fork for now to resolve this. We'll just wait until this can be merged and tagged to pin back.

essen:
Thanks. Please send a patch against Cowlib master to make it apply to future releases. There is no 1.1 release planned at the moment (this will be merged if there is one later on).

nalundgaard:
@essen Isn't that what ninenines/cowlib#54 does?

essen:
I should have looked. :-) Thanks!

essen:
Closing 1.x tickets that are irrelevant for 2.0. Thanks!

PR #1141, opened 2017-03-15
Add inactivity_timeout option for cowboy_http2

Gazler:
The inactivity timeout is used to close the connection in the absence of
any data from the client.

Since this is not part or the rfc7540 spec, a new http2_SUITE module has
been created with a test for the inactivity_timeout.


essen:
I have completely forgotten about this PR when I was working on timeouts. You do have a test though so let me cherry-pick it. :-)

essen:
I have merged the test locally. Will push later. Thanks!

essen:
I am considering dropping this timeout. I expect the `idle_timeout` to trigger first and close the connection. For Websocket we don't have an `inactivity_timeout` and no problems were ever reported.

essen:
I have marked the `inactivity_timeout` as deprecated.

PR #1142, opened 2017-03-15
Handle the Accept header that violates the rfc

define-null:
For accept header that do not follow the rfc https://tools.ietf.org/html/rfc7231, for instance that looks like:
`"Accept: text/html, application/xhtml+xml, application/xml; text/vnd.wap.wml; q=0.9, */*; q=0.8"`

`cowboy_http` would crash with function clause. It's better to return 400 instead.


essen:
Please send a patch against master if necessary. I have no further 1.0 releases planned.

The relevant section of the RFC is https://tools.ietf.org/html/rfc7230#section-3.5 - last paragraph.

Thanks!

define-null:
> I have no further 1.0 releases planned.
Sad to hear that especially for bugfixes.

essen:
There may be further releases done but at customers request only. I only have so much time.

PR #1150, opened 2017-04-06
Update Websocket handler docs default timeout

Gazler:
This was changed from infinity to 60s in
a45813c60f0f983a24ea29d491b37f0590fdd087


essen:
Merged locally, pushed later, thanks!

PR #1154, opened 2017-04-17
Fixed spec of cowboy_req:reply/4 function

mbystryantsev:


jwheare:
I wrote a more complete version of this that fixes duplication in the #http_req record spec and also accepts chunked body funs: https://github.com/ninenines/cowboy/compare/1.0.x...jwheare:1.0.x-dialyzer-fixes

However this is unlikely to be merged. From @essen on IRC:
> there's not going to be more 1.x releases

essen:
Closing 1.x tickets that are irrelevant for 2.0. Thanks!

PR #1155, opened 2017-04-20
Req Body length can be a non_neg_integer or infinity

stevepm:
See docs: https://ninenines.eu/docs/en/cowboy/2.0/guide/req_body/


wk8:
👍 needed here too

essen:
Merged locally, pushed later today. Thanks!

PR #1159, opened 2017-05-05
handle the idle_timeout msg

getong:


essen:
Where does this message come from? It's not something Websocket handlers are supposed to receive/handle. Might be a bug in recent changes. I'll look.

getong:
@essen  I open the websocket, and not send msg for a while and the connection breaks. In the console, the msg:
```
Ranch listener http had connection process started with cowboy_clear:start_link/4 at <0.450.0> exit with reason: {badarg,[
{erlang,iolist_size,[idle_timeout],[]},
{cow_ws,payload_length,1,[{file,"src/cow_ws.erl"},{line,594}]},
{cow_ws,frame,2,[{file,"src/cow_ws.erl"},{line,537}]},
{cowboy_websocket,websocket_send,2,[{file,"src/cowboy_websocket.erl"},{line,368}]},
{cowboy_websocket,handler_call,6,[{file,"src/cowboy_websocket.erl"},{line,338}]},
{cowboy_http,loop,2,[{file,"src/cowboy_http.erl"},{line,180}]},
{cowboy_clear,proc_lib_hack,5,[{file,"src/cowboy_clear.erl"},{line,29}]},
{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,247}]}]}
```
And finally I get this msg out and handle this timeout msg.

essen:
Yep I will write a proper fix, thanks.

getong:
@essen I find the msg where is come from.
The msg is originated from  `cowboy_http.erl`  line 214.
I open the url `http://127.0.0.1:8080/` as the `README.asciidoc` file said, and enter a few words in the web page, and after a while, the msg `{timeout,#Ref<0.0.1.345>, idle_timeout}` come. 
Then this msg come to `cowboy_websocket.erl` function `handler_loop/3`  line 235, and pass the msg to the `ws_handler.erl` file. And crash it.

essen:
Yes but the timer should be canceled, which leads me to think that somewhere the state is ignored. I'll figure it out, don't worry, just finishing my current task. :-)

essen:
I have fixed it in the most recent commit. Thanks!

Issue #1164, opened 2017-05-17
Allow to use If-None-Match HTTP header with PUT method


manifest:
From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match):
> For other methods, and in particular for PUT, If-None-Match used with the * value can be used to save a file not known to exist, guaranteeing that another upload didn't happen before, losing the data of the previous put; this problem is a variation of the [lost update problem](https://www.w3.org/1999/04/Editing/#3.1).

manifest:
and [PR](https://github.com/ninenines/cowboy/pull/1165)

PR #1165, opened 2017-05-17
Allow to use If-None-Match HTTP header with PUT method

manifest:
PR for the [issue](https://github.com/ninenines/cowboy/issues/1164). The patch is pretty straightforward. And we already have [test](https://github.com/ninenines/cowboy/blob/master/test/static_handler_SUITE.erl#L609-L614) for GET/HEAD methods. Do we need any for PUT?

### DIFF
--- a/src/cowboy_rest.erl
+++ b/src/cowboy_rest.erl
@@ -734,7 +734,7 @@ if_none_match_exists(Req, State) ->
 		undefined ->
 			if_modified_since_exists(Req, State);
 		'*' ->
-			precondition_is_head_get(Req, State);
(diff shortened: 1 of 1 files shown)


essen:
The relevant text is https://tools.ietf.org/html/rfc7232#section-3.2

Yes a test would be good. Since it's part of RFC 7232 it should go in a new rfc7232_SUITE module.

manifest:
Ok. Will do.

manifest:
Thanks for pointing to the right spec. Since a server must respond with 412 (Precondition Failed) status code on false condition for PUT method, the PR isn't correct.

PR #1186, opened 2017-07-06
Add a stacktrace to the error message when a callback crashes

rvangraan:
Here is a PR that improves the level of detail logged by Cowboy when a stream handler crashes. The errors used to look like this:
screen shot 2017-07-06 at 12 14 19
It is basically impossible to tell why the handler crashed - look at the yellow rectangle.
After these changes it will look like this:
screen shot 2017-07-06 at 12 55 40


essen:
I've done this a bit differently by adding a report function to cowboy_stream that is called for each callback error. In the future I'm thinking of making this report function configurable. Closing, thanks!

PR #1190, opened 2017-07-26
Example for server push usage

Gazler:
This example uses SSL as it is required in most browsers for server
push.


essen:
I'll wait until 103+cache digests are standardized and implemented to provide a better example of push because it's not very useful at the moment.

essen:
Server push is no longer enabled in Chrome and Firefox, so I don't think we need this example anymore. Server push has been proven unreliable, making things worse than without. The functionality of server push should remain because it may still be useful to APIs and dedicated clients. Closing, thanks!

PR #1081, opened 2017-01-26
max_frame_size option hardcoded to 8 MBytes. Related to issue #612

cybernetlab:

Issue #612, opened 2013-11-10
Allow limiting message size for WebSocket


alco:
Currently, cowboy seems to collect the whole message before invoking the user-provided handler. This poses a security threat if an adversarial client pushes large chunks of data to the server.

There should be a way to specify the maximum message size to prevent memory exhaustion.

essen:
So an idea to solve this fell short: https://twitter.com/MSch/status/561570971069714432

I think what we need first is an option to limit the max size of incoming data (total of all fragments). It should be applied on both fragmented and normal frames and default to a sensible value.

I'm not sure why I said 16MB is the limit for frame size, the maximum size is much much larger.

The default value could be the same value we use for body reading, 8MB.

cybernetlab:
Any progress with it? I think that this is a serious security hole. This allows DDOS by sending a number of huge fragments. WS header has a fragment length field that can be used to drop unwanted packets on early state. I think that there should be something like `max_frag_size` in [cowboy_websocket:state](https://github.com/ninenines/cowboy/blob/master/src/cowboy_websocket.erl#L56) that can be setted to `infinity` by default and would be used to check length [here](https://github.com/ninenines/cowboy/blob/master/src/cowboy_websocket.erl#L234). If it looks applicable I can prepare PR.

Regards

essen:
Was hoping for something useful to come out of max_heap_size but that doesn't include binaries.

I am fine with an option, but it needs to apply to both fragmented and non-fragmented frames (the maximum size for *one* frame is already way too large for most applications). We could name it simply `max_frame_size`. The default should be reasonable, something like 8MB is a good compromise.

cybernetlab:
Can't figure right way to pass this option into state. Should I extend [upgrade](https://github.com/ninenines/cowboy/blob/master/src/cowboy_websocket.erl#L73) with extra argument for this? If so then should I extend `sub_protocol` behaviour - is it breaking change? Other option - use message like `{set_option, max_frame_size, Value}` and catch it in [handler_loop](https://github.com/ninenines/cowboy/blob/master/src/cowboy_websocket.erl#L206). Both approaches doesn't looks good, but the second is more reasonable for me - could you give some hint

essen:
Right, I almost forgot, I have a breaking change to make there to allow passing any number of options. I'd suggest that you start your work with a hardcoded value set at state initialization, and then we can see together. I can probably get it done on Wednesday at most.

cybernetlab:
[...]

PR #1207, opened 2017-08-25
Add max_frame_size option for websocket handlers.

evilbluebeaver:
Option allows to limit a frame by size before decoding its payload.
Related to [612](https://github.com/ninenines/cowboy/issues/612).
I took [this](https://github.com/ninenines/cowboy/pull/1081/commits/2f518e5dbf7d63b7f19f7bb6fe599cb169b183b6) for basis, made it optional and fixed tests. Should I remove limit of 8 MBytes by default or leave it as is?

### DIFF
--- a/src/cowboy_websocket.erl
+++ b/src/cowboy_websocket.erl
@@ -311,12 +311,15 @@ websocket_payload_loop(State=#state{socket=Socket, transport=Transport,
 				end)
 	end.
 
-websocket_dispatch(State=#state{socket=Socket, transport=Transport, frag_state=FragState, frag_buffer=SoFar, extensions=Extensions},
+websocket_dispatch(State=#state{socket=Socket, transport=Transport, frag_state=FragState, frag_buffer=SoFar,
+                                extensions=Extensions, max_frame_size=MaxFrameSize},
 		HandlerState, Type0, Payload0, CloseCode0, RemainingData) ->
 	case cow_ws:make_frame(Type0, Payload0, CloseCode0, FragState) of
 		%% @todo Allow receiving fragments.
 		{fragment, nofin, _, Payload} ->
 			websocket_data(State#state{frag_buffer= << SoFar/binary, Payload/binary >>}, HandlerState, RemainingData);
+        {fragment, fin, _, Payload} when byte_size(Payload) + byte_size(SoFar) > MaxFrameSize ->
+            websocket_close(State, HandlerState, {error, badframe});
 		{fragment, fin, Type, Payload} ->
 			handler_call(State#state{frag_state=undefined, frag_buffer= <<>>}, HandlerState, RemainingData,
 				websocket_handle, {Type, << SoFar/binary, Payload/binary >>}, fun websocket_data/3);
--- a/src/cowboy_websocket.erl
+++ b/src/cowboy_websocket.erl
@@ -316,10 +316,10 @@ websocket_dispatch(State=#state{socket=Socket, transport=Transport, frag_state=F
 		HandlerState, Type0, Payload0, CloseCode0, RemainingData) ->
 	case cow_ws:make_frame(Type0, Payload0, CloseCode0, FragState) of
 		%% @todo Allow receiving fragments.
+        {fragment, _, _, Payload} when byte_size(Payload) + byte_size(SoFar) > MaxFrameSize ->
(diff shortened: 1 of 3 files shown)


essen:
There's another case which you don't cover, which is fragmented frames. We reassemble them and that can end up creating a huge binary too.

evilbluebeaver:
I'm not that deep into websockets protocol. What should I do to check this out?

essen:
It's done in websocket_dispatch, we reassemble the frame from its fragments, should check the size of the (partially or completely) reassembled payload in both clauses.

evilbluebeaver:
Is that ok, that a check performs only on the last fragment of the frame? Or should I add checks on intermediate fragments?

essen:
Yes like this is good. Maybe another fragment test can be done so that it fails on a frame other than the final frame too.

Will have to think about the default, and whether badframe is the right close code, but busy at the moment. Thanks!

lazercorn:
When will this be merged into master?

essen:
It's missing a test as I pointed out in my previous comment.

lazercorn:
So when will it be added?

essen:
You seem to have a lot of time, why don't you do it?

lazercorn:
`
ws_max_frame_size_fragments_close(Config) ->
 	doc("Server closes connection when fragmented frame size exceeds max_frame_size option"),
 	%% max_frame_size is set to 8 bytes in ws_max_frame_size
 	{ok, Socket, _} = do_handshake("/ws_max_frame_size", Config),
 	Mask = 16#11223344,
 	MaskedHello = do_mask(<<"HelloHello">>, Mask, <<>>),
   %% Fragments with exceeding size are not allowed too
 	ok = gen_tcp:send(Socket, << 0:1, 0:3, 2:4, 1:1, 10:7, Mask:32, MaskedHello/binary >>),
 	ok = gen_tcp:send(Socket, << 1:1, 0:3, 0:4, 1:1, 10:7, Mask:32, MaskedHello/binary >>),
	{ok, << 1:1, 0:3, 8:4, 0:1, 2:7, 1002:16 >>} = gen_tcp:recv(Socket, 0, 6000),
 	{error, closed} = gen_tcp:recv(Socket, 0, 6000),
`

What exactly is wrong? 

"Maybe another fragment test can be done so that it fails on a frame other than the final frame too." 
It's failing on the first fragment, not the final.

It's time to merge buddy.

lazercorn:
@evilbluebeaver You mean 0.8 MB not 8 MB?

essen:
It's testing on the last fragment. We need a test for an intermediate fragment. Get to work buddy.

essen:
If you mean what you posted above, I did not read it and I do not intend to test it unless you submit a proper PR. Anyway what's your problem? If you have the code you need in a PR, all you need is to merge it to your copy. Stop bothering maintainers over silly things like this. I don't have plans to go over Websocket for a while so unless there's a PR I can merge without doing extra work it's not likely to be added anytime soon. The test is just one thing that's missing, it's also missing documentation. It's not in a state that can be merged.

lazercorn:
Are you blind? Answer me. Because if you weren't, you'd realize the code I posted isn't testing on the fragment AND it's the exact code as the one the pull requested has, copy pasted.

essen:
> If you mean what you posted above, I did not read it and I do not intend to test it unless you submit a proper PR.

You can either open a PR complete with the added test and documentation, or not do anything. Either way please stop commenting on this PR or I'll lock it. I have better things to do.

lazercorn:
What is honestly your problem? The test that this PR includes two tests. The second one includes two fragments. Both 10 Bytes. Max is 8 Bytes per frame. It's FAILING on the FIRST fragment. Not the LAST. 

10 > 8.

You can either counter my claim or you can lock this PR to admit your ignorance.

essen:
The PR has two tests. One is a non-fragmented frame of 10 bytes that goes over the limit. The second one is two fragmented frames of 5 bytes and it goes over the limit on the last frame.

I'm not sure where you see a test with two fragments of 10 bytes. Maybe from a previous commit, but certainly not from the PR in this current state.

I need a test with say 3 fragments of 5 bytes and it goes over the limit on a frame that's not the last. And I need documentation.

essen:
Added the test, fixed various small things, changed the close code, added docs, etc. and merged locally. Thanks!

PR #1253, opened 2018-01-12
If encountering a wildcard accept-charset match anything

psytale:
Couldn't find any tests for the cowboy rest stuff so haven't added one for this.

Previously if you specify `Accept-Charset: *` and `charsets_provided` is defined it would return a 406


essen:
Fixed in my commit adding tests and fixing a couple more things. Push soon, thanks!

PR #1254, opened 2018-01-15
Improved cowboy_rest charset matching

### LINKED ISSUE #1245: Cowboy_rest issues accepting single charset with content_type_provided/charsets_provided
Appears in 1.* and 2.* to the best of my knowledge.

I have a cowboy_rest endpoint which I want to only provide a single charset. So I expect clients to be specifying their wanted content-type with a charset as a parameter or using the `accept-charset` header. E.g. `accept: application/json;charset=utf-8` or `accept: application/json accept-charset: utf-8`.

The following code does that however it results in a duplicate `charset=utf-8` if they request `application/json;charset=utf-8`
```
content_types_provided(Req, Ctx) ->
  { [ {{<<"application">>, <<"json">>, []}, process_get}
    , {{<<"application">>, <<"json">>, [{<<"charset">>,<<"utf-8">>}]}, process_get}]
  , Req, Ctx}.

charsets_provided(Req, Ctx) ->
  {[<<"utf-8">>], Req, Ctx}.
```
Alternatively I can provide only `{<<"application">>, <<"json">>, '*'}` and always get the correct content-type in the response but a client could request `application/json;charset=ISO-8859-1` and suddenly get a response in the wrong charset instead of a 406 which is not desired.

psytale:
If we match a charset from the accept header we skip the
charsets_provided callback.
If we match a charset in the accept header against a '*' then we use
that against the charsets_provided callback instead of the
accept-charsets header.
Also changed the auto appending of charsets to only occur if its a text
media type and there is not already a charset in the params.

closes #1245

### DIFF
--- a/src/cowboy_rest.erl
+++ b/src/cowboy_rest.erl
@@ -500,18 +500,19 @@ match_media_type(Req, State, Accept, [_Any|Tail], MediaType) ->
 	match_media_type(Req, State, Accept, Tail, MediaType).
 
 match_media_type_params(Req, State, _Accept,
-		[Provided = {{TP, STP, '*'}, _Fun}|_Tail],
+		[{{TP, STP, '*'}, Fun}|_Tail],
 		{{_TA, _STA, Params_A}, _QA, _APA}) ->
 	PMT = {TP, STP, Params_A},
 	languages_provided(Req#{media_type => PMT},
-		State#state{content_type_a=Provided});
+		State#state{content_type_a={{TP, STP, []}, Fun}});
 match_media_type_params(Req, State, Accept,
 		[Provided = {PMT = {_TP, _STP, Params_P}, _Fun}|Tail],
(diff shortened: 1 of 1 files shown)


essen:
I'm happy to say that after getting back to this I still agree with my assessment. I'm now writing tests so that this PR can be merged.

essen:
Alternatively we could accept with `'*'` if there's no `charsets_provided`, and users can use `charsets_provided` to have finer grained control. That's probably more sane than rejecting when it looks like it should be accepted.

essen:
Going for that last one. Of course when that happens we act as if the charset doesn't exist and don't send it back in the response, to avoid sending a potentially invalid charset (we already do that for all params so no change required there).

essen:
This was done on my local copy. Push later today. Thanks for the help figuring this out!

PR #1266, opened 2018-02-17
Allow compression threshold to be configured

### LINKED ISSUE #1265: Allow response body of any size to be compressed
Hi @essen,

I've been looking for  a way to compress all responses, regardless of their size. I noticed that this is currently hardcoded in Cowboy [1](https://github.com/ninenines/cowboy/blob/1.1.x/src/cowboy_req.erl#L899) and [2](https://github.com/ninenines/cowboy/blob/master/src/cowboy_compress_h.erl#L112).

In Cowboy 2 there is [a comment](https://github.com/ninenines/cowboy/blob/master/src/cowboy_compress_h.erl#L104) to maybe make this configurable. I'd be happy to work my way around the codebase to make this happen if:

1/ that's still something you believe valuable
2/ implementation is not too difficult for a newcomer

Thanks

Cowboy version: 1.1.2
(Used via Phoenix 1.3.0)

stevedomin:
Currently the compression threshold is set to 300 and hardcoded in the
codebase. There are cases where it make sense to allow this to be
configured, for instance when you want to enforce all responses to be
compressed regarldess of their size.

Closes #1265

### DIFF
--- a/src/cowboy_compress_h.erl
+++ b/src/cowboy_compress_h.erl
@@ -24,15 +24,20 @@
 -record(state, {
 	next :: any(),
 	compress = undefined :: undefined | gzip,
-	deflate = undefined :: undefined | zlib:zstream()
+	deflate = undefined :: undefined | zlib:zstream(),
+	resp_compression_threshold :: non_neg_integer()
 }).
 
 -spec init(cowboy_stream:streamid(), cowboy_req:req(), cowboy:opts())
 	-> {cowboy_stream:commands(), #state{}}.
 init(StreamID, Req, Opts) ->
 	State0 = check_req(Req),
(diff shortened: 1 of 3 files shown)


stevedomin:
@essen I would love your help figuring out how I can set a configuration value for the test suite.

My understanding is that `cowboy_test:init_common_groups/3` does not take Cowboy config, is that correct?

essen:
For tests with specific configuration you need to create a separate listener. For example: https://github.com/ninenines/cowboy/blob/master/test/http2_SUITE.erl#L48

essen:
I've merged this with some changes. I don't think there needs to be further tests (could be one where it's configured to something else I suppose but that's not terribly important for the moment). Will be pushed soon. Thanks!

stevedomin:
Thanks @essen, never got around to writing the tests for this!

Issue #1234, opened 2017-10-30
[cowboy_http] I try to set `idle_timeout` to 'infinity'


fimars:
Like this, I use on the elixir.
```elixir
    { :ok, _ } = :cowboy.start_clear(:dev_files_server,
      [{ :port, port }],
      %{ :env => %{ :dispatch => build_dispath_config() }, :idle_timeout => :infinity }
    )
```
![image](https://user-images.githubusercontent.com/16299018/32180398-d75b5044-bd5f-11e7-8a57-78608bdcb048.png)

## More
Use 6_000_000`integer` work very well. But i find `idle_timeout` is `timeout()` type.

essen:
I confirm it's missing a case for infinity.

fimars:
I think you're right, thanks!

essen:
I mean it's a bug so leave open if you want it fixed eventually. :-)

fimars:
My fault, I just misunderstood what you mean. 😆  thank you again.

essen:
This should be fixed in Cowboy master, thanks!

PR #1268, opened 2018-03-05
Add case for handling infinity for idle/request_timeout

bwalkowi:
Currently cowboy assumes that idle_timeout or request_timeout is
a number and always starts timers. This commit adds case for
handling infinity as a timeout, allowing to not start mentioned
timers, solving https://github.com/ninenines/cowboy/issues/1234.

### DIFF
--- a/src/cowboy_http.erl
+++ b/src/cowboy_http.erl
@@ -227,8 +227,10 @@ set_timeout(State0=#state{opts=Opts, streams=Streams}) ->
 		[] -> {request_timeout, 5000};
 		_ -> {idle_timeout, 60000}
 	end,
-	Timeout = maps:get(Name, Opts, Default),
-	TimerRef = erlang:start_timer(Timeout, self(), Name),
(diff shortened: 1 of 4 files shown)


essen:
Thanks be we need a test, and we probably need the same to be done for cowboy_http2.

bwalkowi:
I have added case for handling infinity for preface_timeout in http2. As for tests, I can't come up with good idea for tests checking possibility of setting infinity as a timeout ;/

essen:
Just a test that sets the timeout to infinity and checks that the listener still starts. Gun has a few tests in that spirit, for example https://github.com/ninenines/gun/blob/master/test/gun_SUITE.erl#L35

bwalkowi:
Ok, thanks. I will try to write them today or at the beginning of next week at the latest.

bwalkowi:
Ok, I have added tests.

essen:
That was perfect, thanks! This was merged.

PR #1273, opened 2018-03-21
Remove duplicate keys from map in inform docs

Gazler:
This is the equivalent way of doing the same thing using a single header.


essen:
Merged locally, thanks!

PR #1275, opened 2018-03-27
support ranch NbAcceptors option

getong:


essen:
It's supported, it's become the `num_acceptors` transport option.

PR #1296, opened 2018-06-05
Temporary patch for empty host for AWS ALB’s healthcheck

saxxi:
I understand this is not a valid Pull Request and *should not be merged*.

If anyone thinks this is helpful please help us turn it into a configurable option we'd be happy to collaborate.

### DIFF
--- a/src/cowboy_protocol.erl
+++ b/src/cowboy_protocol.erl
@@ -361,7 +361,13 @@ parse_hd_value(<<>>, S, M, P, Q, V, H, N, SoFar) ->
 request(B, State=#state{transport=Transport}, M, P, Q, Version, Headers) ->
 	case lists:keyfind(<<"host">>, 1, Headers) of
 		false when Version =:= 'HTTP/1.1' ->
-			error_terminate(400, State);
+			%% Advenias:
+			%% Why: AWS Application Load Balancer's health check seems to respond with empty `Host:`
+			%% - https://github.com/phoenixframework/phoenix/issues/2437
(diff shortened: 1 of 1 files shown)


src/cowboy_protocol.erl
```diff
@@ -361,7 +361,13 @@ parse_hd_value(<<>>, S, M, P, Q, V, H, N, SoFar) ->
 request(B, State=#state{transport=Transport}, M, P, Q, Version, Headers) ->
 	case lists:keyfind(<<"host">>, 1, Headers) of
 		false when Version =:= 'HTTP/1.1' ->
-			error_terminate(400, State);
+			%% Advenias:
+			%% Why: AWS Application Load Balancer's health check seems to respond with empty `Host:`
```
OvermindDL1: Not 'empty' Host, but rather 'missing' Host, big differences between the two as one follows the spec and the other does not.

essen:
Is it still necessary? This page here https://aws.amazon.com/premiumsupport/knowledge-center/elb-fix-failing-health-checks-alb/ mentions the host header being set to the IP address, which Cowboy would gladly accept (though you might need a catch-all in the router).

essen:
Alternatively it seems you can configure the "Matcher" to consider 400 responses as successful: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html

But I don't know this stuff so maybe I'm looking in the wrong area.

saxxi:
I think we can close this issue, thanks

PR #1308, opened 2018-07-19
Clarify role of DEP_PLUGINS in getting started docs.

robinchew:


essen:
I've made a slightly different edit and will push it soon. Thanks!

PR #1331, opened 2018-10-20
update metrics stream handler to handle `error_response` command

### LINKED ISSUE #1330: metrics stream handler doesn't understand `error_response` command
When cowboy env has 2 stream handlers configured (`[cowboy_metrics_h, cowboy_stream_h]`) and unexpected error happens in one of request handlers, `cowboy_http` [receives](https://github.com/ninenines/cowboy/blob/c998673eb009da2ea4dc0e6ef0332534cf679cc4/src/cowboy_http.erl#L220) an 'EXIT' message, which is [translated](https://github.com/ninenines/cowboy/blob/c998673eb009da2ea4dc0e6ef0332534cf679cc4/src/cowboy_stream_h.erl#L147-L162) into the following chain of commands by `cowboy_stream_h` stream handler (example):

``` erlang 
[{error_response,500,#{<<"content-length">> => <<"0">>},<<>>},
 {log,error,
     "Ranch listener ~p, connection process ~p, stream ~p had its request process ~p exit with reason ~999999p and stacktrace ~999999p~n",
     [bower_app,<0.366.0>,1,<0.367.0>,
      {badmatch,2},
      [<STACKTRACE>]]},
 {internal_error,
     {'EXIT',<0.367.0>,
         {{badmatch,2},
          [<STACKTRACE>]}},
     'Stream process crashed.'}]
```

`cowboy_metrics_h` [does not understand](https://github.com/ninenines/cowboy/blob/c998673eb009da2ea4dc0e6ef0332534cf679cc4/src/cowboy_metrics_h.erl#L196-L245) `error_respone` and does not update state with the response information as it [does](https://github.com/ninenines/cowboy/blob/c998673eb009da2ea4dc0e6ef0332534cf679cc4/src/cowboy_metrics_h.erl#L210) for `response` command.

This results in undefined information about response on terminate step.

velimir0xff:
fixes #1330

### DIFF
--- a/src/cowboy_metrics_h.erl
+++ b/src/cowboy_metrics_h.erl
@@ -207,19 +207,12 @@ fold([{inform, Status, Headers}|Tail],
 		headers => Headers,
 		time => Time
 	}|Infos]});
-fold([{response, Status, Headers, Body}|Tail],
-		State=#state{resp_headers_filter=RespHeadersFilter}) ->
+fold([{response, Status, Headers, Body}|Tail], State) ->
 	Resp = erlang:monotonic_time(),
-	fold(Tail, State#state{
-		resp_status=Status,
-		resp_headers=case RespHeadersFilter of
-			undefined -> Headers;
-			_ -> RespHeadersFilter(Headers)
-		end,
-		resp_start=Resp,
-		resp_end=Resp,
-		resp_body_length=resp_body_length(Body)
-	});
+	fold(Tail, update_with_response(Resp, Status, Headers, Body, State));
+fold([{error_response, Status, Headers, Body}|Tail], State) ->
+	Resp = erlang:monotonic_time(),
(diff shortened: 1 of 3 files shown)


velimir0xff:
@essen is there anything I can do to see reasons of failure? PR builders links are broken for me (page not found).

essen:
Don't think you broke something, it's just some OTP versions and/or tests being more broken than others: https://builds.ninenines.eu/logs/cowboy-prs/92/ubuntu/

essen:
That's not exactly correct. `error_response` is only sent when no other response was sent before. So the metrics handler needs to keep track of that in order to know whether to handle or ignore `error_response`.

essen:
I'll steal the test and fix it now.

velimir0xff:
@essen thank you!

essen:
Basically done locally, push soon.

PR #1335, opened 2018-11-07
Fix ".."/back navigation on directory listing for directory_h.erl

ruanpienaar:
Hi

1) Fix navigating back by using ".." on the directory listing page
2) Fix duplicate addition of "/" to path

A code review would be awesome


essen:
A bit late, sorry. I had a rough year and time just flew.

The solution proposed is not correct because `lists:delete` will remove the first element from the list matching, so if you have a path like `/test/path/test` then you'll end up with `/path/test`.

I'll do a proper fix and merge. Thanks!

essen:
The commit will be 094387a08f95c91ec6e297aa5e5786aff0ab0a65 when I push later today.

PR #1342, opened 2018-12-05
Fixed spec for cowboy_req:reply/4

iantropov:
Hi Everybody,

I found that `cowboy_req:reply/4` has invalid spec (seems like some typo). As it should support all allowed types of `#http_req.resp_body`, I added missed variants.

Without this change Dialyzer warns about mismatches like :

```
The call cowboy_req:reply(200,Req2::cowboy_req:req()) contains an opaque term as 2nd argument when terms of different types are expected in these positions
```


essen:
It's possible however there are no more planned 1.x releases so this will not be merged. Leaving open for a few days before closing to give a chance for other people to see this patch. Thanks!

essen:
I think we can close it now. Thanks!

PR #1351, opened 2019-01-25
idle_timeout increased to 1 hour instead of 60 seconds

amar1994:


essen:
The default is fine. You can configure it in your own application.

amar1994:
Can you please tell how to configure it with other application which is using cowboy?

essen:
Please be more specific.

amar1994:
You said I can configure it in my own application, instead of changing it in cowboy repo.

essen:
Set the protocol options when you start the listener (alongside `env`). The list of options for HTTP/1.1 is there: https://ninenines.eu/docs/en/cowboy/2.6/manual/cowboy_http/#_options

amar1994:
@essen 
Hi,
Passing protocol option (idle_timeout) in **cowboy:start_clear/3** only sets value of idle_timeout in **cowboy_http.erl** module not in cowboy_websocket. And We are using 2.4.0 version of cowboy.

essen:
If Websocket then it's in the options you can return from `init/2`. It's all documented. Not sure what version this got added, you might need to upgrade.

amar1994:
@essen  Thanks.

PR #1353, opened 2019-02-14
Update static_files.asciidoc

tom335:
Minor update clarifying the need of the `priv` directory on project's root.


essen:
Merged locally with tweaks. Thanks!

Issue #1360, opened 2019-03-13
Stacktrace output from cowboy_handler


srstrong:
Currently, the exception handler in cowboy_handler is:

`
  catch Class:Reason ->
    terminate({crash, Class, Reason}, Req, HandlerOpts, Handler),
    erlang:raise(Class, Reason, erlang:get_stacktrace())
  end.
`

If terminate() itself raises but catches an exception, then it is that exception whose stacktrace will get reported.  I discovered this since in my terminate handlers, I often trace out additional debug information - I've just switched over to the new logger, which formats the log in the current process and hits various handled exceptions in libraries such as io_lib_pretty.  This unfortunately results in errors such as:

`
 crasher:
    initial call: cowboy_stream_h:request_process/3
    pid: <0.441.0>
    registered_name: []
    exception error: no match of right hand side value 0
      in function  io_lib_pretty:cind_tag_tuple/7 (io_lib_pretty.erl, line 981)
      in call from io_lib_pretty:while_fail/3 (io_lib_pretty.erl, line 1128)
      in call from io_lib_pretty:print/9 (io_lib_pretty.erl, line 141)
      in call from io_lib_format:build_limited/5 (io_lib_format.erl, line 282)
      in call from io_lib_format:build_limited/5 (io_lib_format.erl, line 302)
      in call from io_lib_format:build_limited/5 (io_lib_format.erl, line 293)
      in call from io_lib_format:build_limited/5 (io_lib_format.erl, line 302)
      in call from io_lib_format:build_limited/5 (io_lib_format.erl, line 293)
    ancestors: [<0.440.0>,<0.248.0>,<0.247.0>,ranch_sup,<0.163.0>]
`

Either the results of erlang:get_stacktrace() should be bound to a variable before calling terminate, or the new pattern-matched approach for getting the stacktrace could be used (although that would prevent older erlang versions)

essen:
Right.

essen:
Done, thanks!

PR #1361, opened 2019-03-13
Bind erlang:get_stacktrace prior to making other calls

srstrong:
Pull request re: [https://github.com/ninenines/cowboy/issues/1360](url)


essen:
Merged locally, thanks!

PR #1364, opened 2019-04-02
Fix calling convention for set_resp_cookie/3

jlouis:
If not providing optional values, they are set to `#{}` and are the last parameter according to the source code. Reflect this in the documentation.


essen:
Merged locally, thanks!

Issue #1369, opened 2019-04-12
How to disable the flow control or keep maintaining the same window size easily?


colachengexo:
Hi author!

We have a component with Cowboy 2.6.1 for internal communication through http2. The client and server are in the same machine. The problem is, client-side will reuse the connection (Gun as the client) in the pool and sometime later we will get this error `{connection_error,flow_control_error,'The flow control window must not be greater than 2^31-1. (RFC7540 6.9.1)'}` in the client.

After more investigation about this, I found the default cowboy_stream_h will not send the window_update immediately instead of sending back window_update after a while the amount according to the length(default 8000000) option in the cowboy_req:read_body_opts(). It will cause the flow_control window on the upward trend until reach 2^31-1 then get the error.  I think because of this [function clause](https://github.com/ninenines/cowboy/blob/2.6.1/src/cowboy_stream_h.erl#L185) in the default stream handler.

I found the elixir grpc library have this [grpc_stream_h.erl](https://github.com/elixir-grpc/grpc/blob/master/src/grpc_stream_h.erl) implementation. It will maintain the same window. But the code base seems old (like Cowboy 2.4).

Do you have any suggestion about this or any configuration option in Cowboy or Gun can tune this?

Thanks!

essen:
I think the problem is rather that Cowboy is naively sending the length here: https://github.com/ninenines/cowboy/blob/2.6.1/src/cowboy_http2.erl#L564

It should just ensure that there's a large enough window and send the difference `Length -CurrentWindow` (if it's positive) instead, I think.

colachengexo:
Sorry, I am not sure your meaning. You mean prevent Cowboy will not send the size to let client window size over the limit? Could you explain more about it? Thanks.

Another question is when this case happens (read_body_length=auto)?
https://github.com/ninenines/cowboy/blob/2.6.1/src/cowboy_stream_h.erl#L96
Because of I saw this case will prepare `{flow, byte_size(Data)}` it means the client will receive window_update immediately. But I will not hit this clause at all.

By the way, do you think this custom stream handler is ok to deal with this problem?

essen:
[...]

PR #1370, opened 2019-04-12
Only send the window_update when Size - RemoteWindow > 0 to prevent over 2^31-1 limit.

colachengexo:
ref: https://github.com/ninenines/cowboy/issues/1369

### DIFF
--- a/src/cowboy_http2.erl
+++ b/src/cowboy_http2.erl
@@ -574,12 +574,18 @@ commands(State0=#state{socket=Socket, transport=Transport, http2_machine=HTTP2Ma
 	commands(State, StreamID, Tail);
 commands(State=#state{socket=Socket, transport=Transport, http2_machine=HTTP2Machine0},
 		StreamID, [{flow, Size}|Tail]) ->
-	Transport:send(Socket, [
-		cow_http2:window_update(Size),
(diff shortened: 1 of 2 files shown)


essen:
Would be good to have a test, not sure how long it'd take to exhaust the window before the patch is applied in a test scenario?

colachengexo:
I wrote a test to let flow window over the limit. But I found the unit test using gun 1.2 won't have this issue because of this version didn't check the window size when receiving the window_update frame. Compare to gun 1.3 https://github.com/ninenines/gun/blob/1.3.0/src/gun_http2.erl#L246. It will cause test cannot finish. I don't know how to write a better test for that. Could you give me some suggestion?
Thanks!

essen:
Yeah I had to downgrade Gun temporarily to release the latest patch. About the test well you should limit the loop of course. Just try to reproduce your issue, count the number of loops, and limit the number of loops to maybe double that if it doesn't take too long.

colachengexo:
I made the limit for the loop. Thanks for your advice.

essen:
Fixed via ninenines/cowboy#1400 and ninenines/cowlib#87. Closing, thanks!

PR #1377, opened 2019-05-20
Doc enhancement: max_frame_size is in unit of bytes.

meetlost:


essen:
Considering there's a lot of sizes not mentioning bytes it would be good to fix them all at once and not just this one.

essen:
Done locally. Thanks!

PR #1382, opened 2019-06-06
Make cowboy compatible with ranch 2

juhlig:
This PR makes cowboy compatible with ranch 2, while it can still be used with ranch 1.


getong:
I think you might also need to update the dep in the `Makefile` and/or the `rebar.config`.

essen:
We are not moving Cowboy to Ranch 2.0 until Cowboy 3.0 is released.

essen:
Merged, thanks!

PR #1392, opened 2019-07-26
Fix lingering_data_frame when data length is 0

tony612:


essen:
Right this must correspond to this that I've seen in Gun:

https://builds.ninenines.eu/logs/gun/133/ubuntu/ct_run.ct_gun@ubuntu0.2019-07-24_14.28.08/ninenines.gun.event_SUITE.logs/run.2019-07-24_14.28.10/unexpected_io.log.html

I don't think the lingering data tuple should be returned at all if the length is 0 to be honest.

essen:
Done by not returning the lingering_data tuple instead. Thanks!

PR #1393, opened 2019-07-30
Use cal_increasing_window to calculate new window

tony612:
Use this PR https://github.com/ninenines/cowlib/pull/83

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ CT_OPTS += -ct_hooks cowboy_ct_hook [] # -boot start_sasl
 LOCAL_DEPS = crypto
 
 DEPS = cowlib ranch
-dep_cowlib = git https://github.com/ninenines/cowlib 2.7.3
+dep_cowlib = git https://github.com/elixir-grpc/cowlib add-new-update-window
 dep_ranch = git https://github.com/ninenines/ranch 1.7.1
 
 DOC_DEPS = asciideck
--- a/src/cowboy_http2.erl
+++ b/src/cowboy_http2.erl
@@ -310,8 +310,7 @@ data_frame(State=#state{opts=Opts, streams=Streams}, StreamID, IsFin, Data) ->
 
 lingering_data_frame(State=#state{socket=Socket, transport=Transport,
 		http2_machine=HTTP2Machine0}, DataLen) ->
-	Transport:send(Socket, cow_http2:window_update(DataLen)),
-	HTTP2Machine1 = cal_and_update_window(DataLen, HTTP2Machine0),
+	HTTP2Machine1 = cal_and_update_window(Transport, Socket, HTTP2Machine0),
 	State#state{http2_machine=HTTP2Machine1}.
 
 headers_frame(State, StreamID, IsFin, Headers,
@@ -581,13 +580,9 @@ commands(State0=#state{socket=Socket, transport=Transport, http2_machine=HTTP2Ma
 	end,
 	commands(State, StreamID, Tail);
 commands(State=#state{socket=Socket, transport=Transport, http2_machine=HTTP2Machine0},
-		StreamID, [{flow, Size}|Tail]) ->
-	Transport:send(Socket, [
-		cow_http2:window_update(Size),
-		cow_http2:window_update(StreamID, Size)
-	]),
-	HTTP2Machine1 = cal_and_update_window(Size, HTTP2Machine0),
-	HTTP2Machine = cal_and_update_window(StreamID, Size, HTTP2Machine1),
+		StreamID, [{flow, _Size}|Tail]) ->
+	HTTP2Machine1 = cal_and_update_window(Transport, Socket, HTTP2Machine0),
+	HTTP2Machine = cal_and_update_window(Transport, Socket, StreamID, HTTP2Machine1),
 	commands(State#state{http2_machine=HTTP2Machine}, StreamID, Tail);
 %% Supervise a child process.
 commands(State=#state{children=Children}, StreamID, [{spawn, Pid, Shutdown}|Tail]) ->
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -15,7 +15,7 @@ CT_OPTS += -ct_hooks cowboy_ct_hook [] # -boot start_sasl
 LOCAL_DEPS = crypto
 
 DEPS = cowlib ranch
-dep_cowlib = git https://github.com/ninenines/cowlib 2.7.3
+dep_cowlib = git https://github.com/elixir-grpc/cowlib add-new-update-window
```
tony612: I'll change this back when the cowlib PR is merged.

src/cowboy_http2.erl
```diff
@@ -586,8 +586,8 @@ commands(State=#state{socket=Socket, transport=Transport, http2_machine=HTTP2Mac
 		cow_http2:window_update(Size),
 		cow_http2:window_update(StreamID, Size)
 	]),
-	HTTP2Machine1 = cow_http2_machine:update_window(Size, HTTP2Machine0),
-	HTTP2Machine = cow_http2_machine:update_window(StreamID, Size, HTTP2Machine1),
+	HTTP2Machine1 = cal_and_update_window(Size, HTTP2Machine0),
+	HTTP2Machine = cal_and_update_window(StreamID, Size, HTTP2Machine1),
```
essen: That can't be right, you send the window update frames with `Size` and then update the state with the calculated size.

tony612:
I don't understand the failure of some tests like
```

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
--
  | http_SUITE:idle_timeout_infinity failed on line 238
  | Reason: {badmatch,#{sock_ip => {127,0,0,1},sock_port => 56431}}
  | - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
```

essen:
Some tests fail because we require an older Gun version for some other tests. Messed up situation, will get fixed whenever I look at it again and move it to Gun master.

src/cowboy_http2.erl
```diff
@@ -666,6 +661,24 @@ maybe_send_data(State=#state{http2_machine=HTTP2Machine0}, StreamID, IsFin, Data
 			send_data(State#state{http2_machine=HTTP2Machine}, SendData)
 	end.
 
+cal_and_update_window(Transport, Socket, HTTP2Machine) ->
```
essen: Just name them `update_window`.

test/rfc7540_SUITE.erl
```diff
@@ -3121,8 +3121,7 @@ lingering_data_counts_toward_connection_window(Config0) ->
 	doc("DATA frames received after sending RST_STREAM must be counted "
 		"toward the connection flow-control window. (RFC7540 5.1)"),
 	Config = cowboy_test:init_http(?FUNCTION_NAME, #{
-		env => #{dispatch => cowboy_router:compile(init_routes(Config0))},
-		initial_connection_window_size => 100000
```
essen: Why change the `initial_connection_window_size` value here? Just curious, if tests pass I have no problem with it.

src/cowboy_http2.erl
```diff
@@ -581,13 +580,9 @@ commands(State0=#state{socket=Socket, transport=Transport, http2_machine=HTTP2Ma
...
 	commands(State, StreamID, Tail);
 commands(State=#state{socket=Socket, transport=Transport, http2_machine=HTTP2Machine0},
-		StreamID, [{flow, Size}|Tail]) ->
-	Transport:send(Socket, [
-		cow_http2:window_update(Size),
-		cow_http2:window_update(StreamID, Size)
-	]),
-	HTTP2Machine1 = cow_http2_machine:update_window(Size, HTTP2Machine0),
-	HTTP2Machine = cow_http2_machine:update_window(StreamID, Size, HTTP2Machine1),
+		StreamID, [{flow, _Size}|Tail]) ->
```
essen: We have to take the size into account. If the user requests for 8MB of data then we should set the window to 8MB directly instead of sending so many `window_update` frames.

essen: I think if the requested flow size is larger than our ideal window we should just send the flow size, otherwise we should use your algorithm in order to avoid sending too many `window_update` frames.

essen: Cowlib should probably have this size as argument and use it as part of the decision.

src/cowboy_http2.erl
tony612: But we already have a update_window, right?

test/rfc7540_SUITE.erl
tony612: I'm fixing this test and found it's not necessary to set custom initial_connection_window_size here. And we need creating a larger binary for tests.

src/cowboy_http2.erl
tony612: OK. I agree with that.

essen:
The more I think about it the more I think this is not the right solution. This is what I'm thinking of right now:

We want to keep two values: the `window` (as we already do) and the requested `flow`. We want to ensure that at all times `flow` + `minimum window size` =< `window`. We keep `minimum window size` in order to have room for data to come in while we are processing what we already received.

The `flow` for the connection is the total between all the streams. It increases when a stream `flow` is increased, and decreases when a stream completes or is canceled.

When we receive a `flow` command we only send a `WINDOW_UPDATE` and update the `window` when the condition above is false. If there is already enough room in the window, we only update the `flow` and do nothing further.

When we receive a `flow` command and we do need to send a `WINDOW_UPDATE`, we calculate how much we actually need for the condition to become false again. However in order to avoid sending too many `WINDOW_UPDATE` frames we update the `window` so that it fits all our `flow` + the `minimum window size`, leaving some extra room again for upcoming frames.

When we receive lingering data we update the `window` but do not send a `WINDOW_UPDATE` as we are already ensuring that the connection `window` is higher than the `window` for all the streams.

This should take care of reducing the number of `WINDOW_UPDATE` we send. In the future we may also want to have an upper bound to the connection `window` but I suggest leaving it out for the time being.

Thoughts?

essen:
I've opened two PRs to add heuristics to fix issues and greatly improve WINDOW_UPDATE handling. Please take a look: https://github.com/ninenines/cowlib/pull/87 #1400 

I will document the new options later today or on Thursday.

essen:
Your original issue has been fixed via ninenines/cowboy#1400 and ninenines/cowlib#87. Closing, thanks!

PR #1399, opened 2019-08-16
Gracefully shut down HTTP/2 connections on GOAWAY

essen:
Minimal set of changes to make the Gun test suite in https://github.com/ninenines/gun/pull/206 to pass. There might be more scenarios where we want a graceful shutdown.

We should also not try to send pushed resources when we are closing.


essen:
Merged this one.

PR #1400, opened 2019-09-02
Fix and optimize sending of WINDOW_UPDATE frames

essen:
For long-running connections it was possible for the connection
window to become larger than allowed by the protocol because the
window increases claimed by stream handlers were never reclaimed
even if no data was consumed.

The new code applies heuristics to fix this and reduce the number
of WINDOW_UPDATE frames that are sent. It includes six new options
to control that behavior: margin, max and threshold for both the
connection and stream windows. The margin is some extra space
added on top of the requested read size. The max is the maximum
window size at any given time. The threshold is a minimum window
size that must be reached before we even consider sending more
WINDOW_UPDATE frames. We also avoid sending WINDOW_UPDATE frames
when there is already enough space in the window, or when the
read size is 0.

Fix for the problems signaled through tickets and PRs https://github.com/ninenines/cowlib/pull/81 https://github.com/ninenines/cowlib/pull/83 https://github.com/ninenines/cowboy/issues/1369 https://github.com/ninenines/cowboy/pull/1370 https://github.com/ninenines/cowboy/pull/1393 and possibly more.

Companion PR https://github.com/ninenines/cowlib/pull/87

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ CT_OPTS += -ct_hooks cowboy_ct_hook [] # -boot start_sasl
 LOCAL_DEPS = crypto
 
 DEPS = cowlib ranch
-dep_cowlib = git https://github.com/ninenines/cowlib 2.7.3
+dep_cowlib = git https://github.com/ninenines/cowlib better-window-updates
 dep_ranch = git https://github.com/ninenines/ranch 1.7.1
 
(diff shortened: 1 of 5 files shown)


essen:
Merge cowlib first, change the branch to master here, and should be good to merge.

essen:
And merged!

Issue #1376, opened 2019-05-08
cowboy_req:stream_body/3 needs a backpressure mechanism


jvoegele:
We have a Phoenix application that generates very large API responses (approaching 1Gb in size) in both JSON and CSV formats. To keep memory usage down, we are streaming these large responses back to the client using `Plug.Conn.send_chunked` and `Plug.Conn.chunk`, which in turn calls `cowboy_req:stream_body/3`.

The problematic behavior that we have experienced with this is that, because `cowboy_req:stream_body/3` does not have any sort of backpressure mechanism, the stream is fully materialized almost immediately and memory usage spikes to hold the entire stream in memory. Furthermore, the `cowboy_clear:connection_process` process gets overloaded with millions of messages in its mailbox, and hence slows down to a crawl.

I created an issue in the GitHub issue tracker for the `plug_cowboy` project, and rather than repeat the lengthy analysis here I will simply refer to that issue for further detail: https://github.com/elixir-plug/plug_cowboy/issues/10

The upshot is that there needs to be some coordination between `Plug.Cowboy.Conn.chunk/2` and `cowboy_req:stream_body/3` to provide backpressure and prevent the spike in memory usage and the extreme performance degradation. @josevalim suggests that Cowboy could provide some sort of ping operation that would return a pong to the caller once it receives the ping from its mailbox. See this link: https://github.com/elixir-plug/plug_cowboy/issues/10#issuecomment-485133346

Does that seem like a reasonable approach, or do you have any other ideas for addressing this issue?

Thanks!

essen:
In Cowboy 1.0 the `chunk` function was synchronous because it was writing directly to the socket. I don't have a problem making the 2.0 equivalent's `stream_body` synchronous as well. The implementation would then be restricted to the `cowboy_stream` related modules.

essen:
As a workaround in the meantime you can always check the size of the mailbox of the connection process before calling stream_body again. Should be cheap.

josevalim:
While I am totally fine with `chunk` being `sync`, it is worth saying the advantage of a `sync`/`ping` operation is that it allows us to be more efficient. Instead of syncing on every write, I can write 10 or 100, and only then wait until that is flushed.

`message_queue_len` unfortunately uses locks, I am not sure of the granularity of the lock though.

essen:
As a workaround I don't think it'd be a big issue to check it every few stream_body calls and sleep between tries. Obviously not what the real solution should look like.

[...]

PR #1402, opened 2019-09-13
Implement backpressure on cowboy_req:stream_body

essen:
This should limit the amount of memory that Cowboy is using
when a handler is sending data much faster than the network.

The new max_stream_buffer_size is a soft limit and only has
an effect when the cowboy_stream_h handler is used.

https://github.com/ninenines/cowboy/issues/1376


essen:
All the changes I wanted should now be in.

essen:
Guess I can merge this.

essen:
Merged!

Issue #1427, opened 2019-12-20
Don't fail when server has {compress, true} enabled and receives a malformed Accept-Encoding header 


mracos:
Regarding this comment here, https://github.com/ninenines/cowboy/blob/63b17e4edf666d995ec86cdcda17a62ba5ebc423/src/cowboy_compress_h.erl#L75

Not sure how to solve this or even if it is still a desirable thing to be done, should we just ignore the header on these situations and not compress the response by maybe wrapping this block in a `try/catch`? Should we return a `406 NOT ACCEPTABLE`? WDYT?

---

Last but not least, thank you for all the awesome work you do ❤️

essen:
Did you run into this in the wild?

The RFC doesn't say but generally we're free to ignore the accept headers so I would suggest doing so in this case and not compressing.

mracos:
So, _Kinda_, hahaha
I had to do something that was exactly what `cowboy_compress_h` already does so in order to understand if I could use I went on to read the source code. 😄

So I'll give this one a shot and come up with a patch, does that sound reasonable?

essen:
Sure. Make sure to leave a comment explaining (since this isn't mentioned in the RFC) and add a test.

Found this by the way: https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.3.4 So error handling really is application specific. I think not throwing an error is the better choice for this particular case.

essen:
Done, thanks.

PR #1429, opened 2019-12-24
Don't raise an error for malformed accept-encoding headers in compressed

mracos:
Related to #1427

### DIFF
--- a/src/cowboy_compress_h.erl
+++ b/src/cowboy_compress_h.erl
@@ -72,8 +72,7 @@ early_error(StreamID, Reason, PartialReq, Resp, Opts) ->
 
 %% Check if the client supports decoding of gzip responses.
 check_req(Req) ->
-	%% @todo Probably shouldn't unconditionally crash on failure.
-	case cowboy_req:parse_header(<<"accept-encoding">>, Req) of
+	try cowboy_req:parse_header(<<"accept-encoding">>, Req) of
 		%% Client doesn't support any compression algorithm.
 		undefined ->
 			#state{compress=undefined};
@@ -87,6 +86,10 @@ check_req(Req) ->
 				_ ->
 					#state{compress=gzip}
 			end
+  catch
+		%% Don't raise an error because of malformed or unsupported accept-encoding headers.
+		exit:{request_error, {header, <<"accept-encoding">>}, _Reason} ->
+			#state{compress=undefined}
 	end.
 
 %% Do not compress responses that contain the content-encoding header.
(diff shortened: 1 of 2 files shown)


src/cowboy_compress_h.erl
```diff
@@ -72,21 +72,26 @@ early_error(StreamID, Reason, PartialReq, Resp, Opts) ->
...
+				%% equivalent to "gzip". (RFC7230 4.2.3)
+				case [E || E={<<"gzip">>, Q} <- Encodings, Q =/= 0] of
+					[] ->
+						#state{compress=undefined};
+					_ ->
+						#state{compress=gzip}
+				end
+		end
+  catch
+		%% Don't raise an error because of malformed or unsupported accept-encoding headers.
```
mracos: I was not sure on what to put here, improvements appreciated 😄

src/cowboy_compress_h.erl
```diff
@@ -72,21 +72,26 @@ early_error(StreamID, Reason, PartialReq, Resp, Opts) ->
...
-				_ ->
-					#state{compress=gzip}
-			end
+	try
+		case cowboy_req:parse_header(<<"accept-encoding">>, Req) of
+			%% Client doesn't support any compression algorithm.
+			undefined ->
+				#state{compress=undefined};
+			Encodings ->
+				%% We only support gzip so look for it specifically.
```
mracos: I thought about doing this one too, wdyt?

src/cowboy_compress_h.erl
```diff
@@ -72,21 +72,26 @@ early_error(StreamID, Reason, PartialReq, Resp, Opts) ->
...
-			%% @todo A recipient SHOULD consider "x-gzip" to be
-			%% equivalent to "gzip". (RFC7230 4.2.3)
-			case [E || E={<<"gzip">>, Q} <- Encodings, Q =/= 0] of
-				[] ->
-					#state{compress=undefined};
-				_ ->
-					#state{compress=gzip}
-			end
+	try
+		case cowboy_req:parse_header(<<"accept-encoding">>, Req) of
```
essen: Don't wrap the case in a try, replace the case with the try. `try cowboy_req:parse... of ... catch`

src/cowboy_compress_h.erl
essen: No don't because that will change eventually.

src/cowboy_compress_h.erl
mracos: Nice one! [done here](https://github.com/ninenines/cowboy/compare/deb6dba33003528e934124c5d0827aa5d68eead7..8ffbe073c294743a5ce77ba2d637697b53306dcb).

essen:
I'll try to get to this later today.

essen:
Tweaked things and merged, thanks!

PR #1432, opened 2019-12-31
Update cowlib

tony612:
This is used to verify https://github.com/ninenines/cowlib/pull/93


essen:
Merged the updated test, thanks!

Issue #1411, opened 2019-10-30
badarith,[{erlang,'+',[infinity,64000]


stofel:
On load testing current master tag 3000rps get crashes
=ERROR REPORT==== 30-Oct-2019::09:55:08.485504 ===
Ranch listener my_http_listener had connection process started with cowboy_clear:start_link/4 at <0.29776.42> exit with reason: {badarith,[{erlang,'+',[infinity,64000],[]},{cowboy_http,commands,3,[{file,"/home/alexs/work/psp/_build/default/lib/cowboy/src/cowboy_http.erl"},{line,948}]},{cowboy_http,loop,1,[{file,"/home/alexs/work/psp/_build/default/lib/cowboy/src/cowboy_http.erl"},{line,234}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,249}]}]}

essen:
It's missing a clause for when a flow (read body) command arrives after the body was finished reading.

stofel:
And?
I use
`
init(Req, Opts) ->
 M = cowboy_req:method(Req),
 B = cowboy_req:read_urlencoded_body(Req)
 ...
`
read_urlencoded_body crashed
Is here need some patch?

essen:
I'm just saying why there's a crash, I'll work on a fix later.

roivinir:
Seeing the same crash as @stofel also under heavy load 
Also a very similar one soon there after. Looks like the same issue.

"Ranch listener ApiServicesWeb.Endpoint.HTTP had connection process started with :cowboy_clear:start_link/4 at #PID<0.19133.8> exit with reason: {:badarith, [{:erlang, :-, [:infinity, 1460], []}, {:cowboy_http, :update_flow, 3, [file: '/home/.../deps/cowboy/src/cowboy_http.erl', line: 381]}, {:cowboy_http, :after_parse, 1, [file: '/home/.../deps/cowboy/src/cowboy_http.erl', line: 358]}, {:proc_lib, :init_p_do_apply, 3, [file: 'proc_lib.erl', line: 249]}]}\n",

essen:
Feel free to experiment with a fix I don't think I'll get to it this week.

essen:
Please try this patch:

``` patch
diff --git a/src/cowboy_http.erl b/src/cowboy_http.erl
index f128a44..d60f4e0 100644
--- a/src/cowboy_http.erl
+++ b/src/cowboy_http.erl
[...]

PR #1434, opened 2020-02-07
fix badarith on cowboy_http:commands

alexandre-kivra:
This PR applies @essen patch of comment https://github.com/ninenines/cowboy/issues/1411#issuecomment-549314493

We have it running for a few days and seems to at least don't cause any other bugs.


essen:
I've started writing a test for this, I have the change otherwise available on my local copy. I'll commit/push when I am done.

Issue #1448, opened 2020-04-18
cowboy router: catchall host and concrete host don't work together


farao:
I am trying to start a websockethandler on all available "hosts" (networks) and an http handler only on localhost. Even though this shouldn't be a contradiction, it always only works for the first entry, the second entry seems to be ignored:
```
:cowboy_router.compile([
  {:_,          [{"/[..]", WebsocketHandler, []}]},
  {"localhost", [{"/metrics", HTTPHandler, []}]}
])
```
lets the websocket handler work but not the http handler (returns 404) and
```
:cowboy_router.compile([
  {"localhost", [{"/metrics", HTTPHandler, []}]}
  {:_,          [{"/", WebsocketHandler, []}]}
])
```
the other way round. In both cases it opens the chosen port on 0.0.0.0 (all hosts).
How can I make both work at the same time?

essen:
The second one is half correct. You however need to specify the Websocket handler in all hosts you want it to work on (so both localhost and `_`), because the host match stops once a match was found.

farao:
Thanks! Could I have found this information somewhere (except in the code)?

essen:
Don't know if it's explicitly stated, please double check.

PR #1449, opened 2020-04-18
Document cowboy_router host/path matching order

farao:
Explaining better how to write routes, coming from this misunderstanding:
https://github.com/ninenines/cowboy/issues/1448


essen:
I have expanded further and will push this soon. Thanks!

PR #1463, opened 2020-08-18
Minor grammar corrections

Thomas-Sciaroni:
Verb agreement error "The request process executes middlewares which, by default, including the router and then the execution of handlers." -> "The request process executes middlewares. By default, the request process executes the router and then the handlers."

adverbial clause at beginning of sentence needs comma "By default Cowboy comes..." -> "By default, Cowboy comes..."


essen:
Merged, thanks!

PR #1464, opened 2020-09-11
Added ret val  `{created|see_other, URI}` from AcceptCallback

mbj4668:


doc/src/manual/cowboy_rest.asciidoc
```diff
@@ -99,10 +103,10 @@ For PUT requests, the body is a representation of the resource
 that is being created or replaced.
 
 For POST requests, the body is typically application-specific
-instructions on how to process the request, but it may also
-be a representation of the resource. When creating a new
-resource with POST at a different location, return `{true, URI}`
-with `URI` the new location.
+instructions on how to process the request, but it may also be a
+representation of the resource. When creating a new resource with POST
+at a different location, return `{create, URI}` or `{see_other, URI`}
```
essen, line 108: `created`

It should probably be explained that the latter redirects while the former does not.

doc/src/manual/cowboy_rest.asciidoc
```diff
@@ -99,10 +103,10 @@ For PUT requests, the body is a representation of the resource
...
+with `URI` the new location. The return value `{true, URI}` is deprecated.
```
essen, line 109: If it's deprecated then it should no longer be documented. Also the deprecation should be indicated in the changelog at the bottom of the file instead.

test/rest_handler_SUITE.erl
```diff
@@ -474,6 +475,39 @@ delete_resource_missing(Config) ->
...
+create_resource_legacy(Config) ->
+	doc("POST to an existing resource to create a new resource. "
+		"When the accept callback returns {true, NewURI}, the legacy "
+		"behavior is a 302 See Other reply."),
+	ConnPid = gun_open(Config),
+	Ref = gun:post(ConnPid, "/create_resource", [
+		{<<"content-type">>, <<"application/text">>}
+	], <<"hello">>, #{}),
+	{response, _, 303, _} = gun:await(ConnPid, Ref),
+	ok.
```
essen, line 509: Do not test the old behavior. On the other hand it would be worth testing `true` and `false` return values if they aren't already.

essen (changes requested):
Thanks for taking this on! I have a few small comments inline.

doc/src/manual/cowboy_rest.asciidoc
mbj4668: Ok.  Which version number should I use?  2.9?

essen: Yes.

test/rest_handler_SUITE.erl
mbj4668: Ok.  I am used to document deprecated "stuff" and keep the tests, and then after N releases the deprecated "stuff" is removed, including the documentation and tests.  I think it is clear, but I will of course adapt to your policy.

`true` and `false` are already tested by `accept_callback_h.erl`

essen: Keep the tests yes, not add new ones. :-)

I only document what should be used, on the other hand I keep the documentation from previous releases with a version select so people can follow whichever they're currently using.

doc/src/manual/cowboy_rest.asciidoc
```diff
@@ -99,11 +102,15 @@ For PUT requests, the body is a representation of the resource
...
-resource with POST at a different location, return `{true, URI}`
+instructions on how to process the request, but it may also be a
+representation of the resource. When creating a new resource with POST
+at a different location, return `{created, URI}` or `{see_other, URI`}
 with `URI` the new location.
 
+Returning `{created, URI`} will result in a '201 Created' response.
+
+Returning `{see_other, URI`} will result in a '303 See Other' response.
+
```
essen, line 113: That's not what I meant. The exact status code is not important here (it will be important in the updated diagrams). What's important is to say that the `see_other` return value implies a redirect. Something like

```
The `see_other` tuple will redirect the client to the new location automatically.
```

at the end of the paragraph.

doc/src/manual/cowboy_rest.asciidoc
```diff
@@ -724,6 +731,9 @@ listed here, like the authorization header.
 
 == Changelog
 
+* *2.9*: An `AcceptCallback` can now return `{created, URI}` and
```
essen, line 734: -and +or

doc/src/manual/cowboy_rest.asciidoc
mbj4668: I tried to follow the style of other callbacks; the first one I found was delete_completed which has:
```
Returning `false` will result in a '202 Accepted' response
```
I will change this to the sentence you suggested.

essen: Right the existing manual isn't very consistent in that regard.

essen:
Merged, thanks!

PR #1471, opened 2020-10-08
Graceful shutdown (HTTP/1.x and HTTP/2)

zuiderkwast:
For HTTP/2:

1. A GOAWAY frame with the last stream id set to 2^31-1 is sent and a
   timer is started (goaway_initial_timeout, default 1000ms), to wait
   for any in-flight requests sent by the client, and the status is set
   to 'closing_initiated'. If the client responds with GOAWAY and closes
   the connection, we're done.
2. A second GOAWAY frame is sent with the actual last stream id and the
   status is set to 'closing'. If no streams exist, the connection
   terminates. Otherwise a second timer (goaway_complete_timeout,
   default 3000ms) is started, to wait for the streams to complete. New
   streams are not accepted when status is 'closing'.
3. If all streams haven't completed after the second timeout, the
   connection is forcefully terminated.

For HTTP/1.x:

1. If a request is currently being handled, it is waited for and the
   response is sent back to the client with the header "Connection:
   close". Then, the connection is closed.
2. If the current request handler is not finished within the time
   configured in transport option 'shutdown' (default 5000ms), the
   connection process is killed by its supervisor (ranch).

Implemented for HTTP/1.x and HTTP/2 in the following scenarios:

* When receiving exit signal 'shutdown' from the supervisor
* When sys:terminate/2,3 is called

### DIFF
--- a/src/cowboy_http.erl
+++ b/src/cowboy_http.erl
@@ -245,6 +245,9 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport, opts=Opts,
 		{timeout, _, _} ->
 			loop(State);
 		%% System messages.
+		{'EXIT', Parent, shutdown} ->
+			Reason = {stop, {exit, shutdown}, 'Parent process requested shutdown.'},
+			loop(initiate_closing(State, Reason));
 		{'EXIT', Parent, Reason} ->
 			terminate(State, {stop, {exit, Reason}, 'Parent process terminated.'});
 		{system, From, Request} ->
@@ -1440,6 +1443,13 @@ early_error(StatusCode0, #state{socket=Socket, transport=Transport,
 	end,
 	ok.
 
+initiate_closing(State=#state{streams=[]}, Reason) ->
+	terminate(State, Reason);
+initiate_closing(State=#state{streams=[_Stream|Streams],
+		out_streamid=OutStreamID}, Reason) ->
+	terminate_all_streams(State, Streams, Reason),
+	State#state{last_streamid = OutStreamID}.
+
 -spec terminate(_, _) -> no_return().
 terminate(undefined, Reason) ->
 	exit({shutdown, Reason});
@@ -1503,9 +1513,10 @@ terminate_linger_loop(State=#state{socket=Socket}, TimerRef, Messages) ->
 system_continue(_, _, State) ->
 	loop(State).
 
--spec system_terminate(any(), _, _, {#state{}, binary()}) -> no_return().
-system_terminate(Reason, _, _, State) ->
-	terminate(State, {stop, {exit, Reason}, 'sys:terminate/2,3 was called.'}).
+-spec system_terminate(any(), _, _, #state{}) -> no_return().
+system_terminate(Reason0, _, _, State) ->
+	Reason = {stop, {exit, Reason0}, 'sys:terminate/2,3 was called.'},
+	loop(initiate_closing(State, Reason)).
 
 -spec system_code_change(Misc, _, _, _) -> {ok, Misc} when Misc::{#state{}, binary()}.
 system_code_change(Misc, _, _, _) ->
--- a/src/cowboy_http2.erl
+++ b/src/cowboy_http2.erl
@@ -31,6 +31,8 @@
 	connection_window_update_threshold => 0..16#7fffffff,
 	enable_connect_protocol => boolean(),
 	env => cowboy_middleware:env(),
+	goaway_initial_timeout => timeout(),
+	goaway_complete_timeout => timeout(),
 	idle_timeout => timeout(),
 	inactivity_timeout => timeout(),
 	initial_connection_window_size => 65535..16#7fffffff,
@@ -88,7 +90,7 @@
 	proxy_header :: undefined | ranch_proxy_header:proxy_info(),
 	opts = #{} :: opts(),
 
-	%% Timer for idle_timeout.
+	%% Timer for idle_timeout; also used for goaway timers.
 	timer = undefined :: undefined | reference(),
 
 	%% Remote address and port for the connection.
@@ -101,7 +103,7 @@
 	cert :: undefined | binary(),
 
 	%% HTTP/2 state machine.
-	http2_status :: sequence | settings | upgrade | connected | closing,
+	http2_status :: sequence | settings | upgrade | connected | closing_initiated | closing,
 	http2_machine :: cow_http2_machine:http2_machine(),
 
 	%% HTTP/2 frame rate flood protection.
@@ -160,7 +162,7 @@ init(Parent, Ref, Socket, Transport, ProxyHeader, Opts) ->
 	binary() | undefined, binary()) -> ok.
 init(Parent, Ref, Socket, Transport, ProxyHeader, Opts, Peer, Sock, Cert, Buffer) ->
 	{ok, Preface, HTTP2Machine} = cow_http2_machine:init(server, Opts),
-	State = set_timeout(init_rate_limiting(#state{parent=Parent, ref=Ref, socket=Socket,
+	State = set_idle_timeout(init_rate_limiting(#state{parent=Parent, ref=Ref, socket=Socket,
 		transport=Transport, proxy_header=ProxyHeader,
 		opts=Opts, peer=Peer, sock=Sock, cert=Cert,
 		http2_status=sequence, http2_machine=HTTP2Machine})),
@@ -205,7 +207,7 @@ init(Parent, Ref, Socket, Transport, ProxyHeader, Opts, Peer, Sock, Cert, Buffer
 		<<"connection">> => <<"Upgrade">>,
 		<<"upgrade">> => <<"h2c">>
 	}, ?MODULE, undefined}), %% @todo undefined or #{}?
-	State = set_timeout(init_rate_limiting(State2#state{http2_status=sequence})),
+	State = set_idle_timeout(init_rate_limiting(State2#state{http2_status=sequence})),
 	Transport:send(Socket, Preface),
 	setopts_active(State),
 	case Buffer of
@@ -227,9 +229,13 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport,
 	receive
 		%% Socket messages.
 		{OK, Socket, Data} when OK =:= element(1, Messages) ->
-			parse(set_timeout(State), << Buffer/binary, Data/binary >>);
+			parse(set_idle_timeout(State), << Buffer/binary, Data/binary >>);
 		{Closed, Socket} when Closed =:= element(2, Messages) ->
-			terminate(State, {socket_error, closed, 'The socket has been closed.'});
+			Reason = case State#state.http2_status of
+				closing -> {stop, closed, 'The client is going away.'};
+				_ -> {socket_error, closed, 'The socket has been closed.'}
+			end,
+			terminate(State, Reason);
 		{Error, Socket, Reason} when Error =:= element(3, Messages) ->
 			terminate(State, {socket_error, Reason, 'An error has occurred on the socket.'});
 		{Passive, Socket} when Passive =:= element(4, Messages);
@@ -238,8 +244,10 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport,
 			setopts_active(State),
 			loop(State, Buffer);
 		%% System messages.
+		{'EXIT', Parent, shutdown} ->
+			Reason = {stop, {exit, shutdown}, 'Parent process requested shutdown.'},
+			loop(initiate_closing(State, Reason), Buffer);
 		{'EXIT', Parent, Reason} ->
-			%% @todo Graceful shutdown here as well?
 			terminate(State, {stop, {exit, Reason}, 'Parent process terminated.'});
 		{system, From, Request} ->
 			sys:handle_system_msg(Request, From, Parent, ?MODULE, [], {State, Buffer});
@@ -252,6 +260,11 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport,
 			loop(State, Buffer);
 		{timeout, TRef, {cow_http2_machine, Name}} ->
 			loop(timeout(State, Name, TRef), Buffer);
+		{timeout, TimerRef, {goaway_initial_timeout, Reason}} ->
+			loop(closing(State, Reason), Buffer);
+		{timeout, TimerRef, {goaway_complete_timeout, Reason}} ->
+			terminate(State, {stop, stop_reason(Reason),
+				'Graceful shutdown timed out.'});
 		%% Messages pertaining to a stream.
 		{{Pid, StreamID}, Msg} when Pid =:= self() ->
 			loop(info(State, StreamID, Msg), Buffer);
@@ -269,14 +282,21 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport,
 		terminate(State, {internal_error, timeout, 'No message or data received before timeout.'})
 	end.
 
-set_timeout(State=#state{opts=Opts, timer=TimerRef0}) ->
+set_idle_timeout(State=#state{http2_status=Status, timer=TimerRef})
+		when Status =:= closing_initiated orelse Status =:= closing,
+			TimerRef =/= undefined ->
+	State;
+set_idle_timeout(State=#state{opts=Opts}) ->
+	set_timeout(State, maps:get(idle_timeout, Opts, 60000), idle_timeout).
+
+set_timeout(State=#state{timer=TimerRef0}, Timeout, Message) ->
 	ok = case TimerRef0 of
 		undefined -> ok;
 		_ -> erlang:cancel_timer(TimerRef0, [{async, true}, {info, false}])
 	end,
-	TimerRef = case maps:get(idle_timeout, Opts, 60000) of
+	TimerRef = case Timeout of
 		infinity -> undefined;
-		Timeout -> erlang:start_timer(Timeout, self(), idle_timeout)
+		Timeout -> erlang:start_timer(Timeout, self(), Message)
 	end,
 	State#state{timer=TimerRef}.
 
@@ -567,18 +587,24 @@ timeout(State=#state{http2_machine=HTTP2Machine0}, Name, TRef) ->
 
 %% Erlang messages.
 
-down(State=#state{opts=Opts, children=Children0}, Pid, Msg) ->
-	case cowboy_children:down(Children0, Pid) of
+down(State0=#state{opts=Opts, children=Children0}, Pid, Msg) ->
+	State = case cowboy_children:down(Children0, Pid) of
 		%% The stream was terminated already.
 		{ok, undefined, Children} ->
-			State#state{children=Children};
+			State0#state{children=Children};
 		%% The stream is still running.
 		{ok, StreamID, Children} ->
-			info(State#state{children=Children}, StreamID, Msg);
+			info(State0#state{children=Children}, StreamID, Msg);
 		%% The process was unknown.
 		error ->
 			cowboy:log(warning, "Received EXIT signal ~p for unknown process ~p.~n",
 				[Msg, Pid], Opts),
+			State0
+	end,
+	if
+		State#state.http2_status =:= closing, State#state.streams =:= #{} ->
+			terminate(State, {stop, normal, 'The connection is going away.'});
+		true ->
 			State
 	end.
 
@@ -909,19 +935,21 @@ stream_alarm(State, StreamID, Name, Value) ->
 %% We may have to cancel streams even if we receive multiple
 %% GOAWAY frames as the LastStreamID value may be lower than
 %% the one previously received.
-goaway(State0=#state{socket=Socket, transport=Transport, http2_machine=HTTP2Machine,
+goaway(State0=#state{socket=Socket, transport=Transport, http2_machine=HTTP2Machine0,
 		http2_status=Status, streams=Streams0}, {goaway, LastStreamID, Reason, _})
-		when Status =:= connected; Status =:= closing ->
+		when Status =:= connected; Status =:= closing_initiated; Status =:= closing ->
 	Streams = goaway_streams(State0, maps:to_list(Streams0), LastStreamID,
 		{stop, {goaway, Reason}, 'The connection is going away.'}, []),
 	State = State0#state{streams=maps:from_list(Streams)},
-	case Status of
-		connected ->
+	if
+		Status =:= connected; Status =:= closing_initiated ->
+			{OurLastStreamID, HTTP2Machine} =
+				cow_http2_machine:set_last_streamid(HTTP2Machine0),
 			Transport:send(Socket, cow_http2:goaway(
-				cow_http2_machine:get_last_streamid(HTTP2Machine),
-				no_error, <<>>)),
-			State#state{http2_status=closing};
-		_ ->
+				OurLastStreamID, no_error, <<>>)),
+			State#state{http2_status=closing,
+				http2_machine=HTTP2Machine};
+		true ->
 			State
 	end;
 %% We terminate the connection immediately if it hasn't fully been initialized.
@@ -938,21 +966,65 @@ goaway_streams(State, [{StreamID, #stream{state=StreamState}}|Tail], LastStreamI
 goaway_streams(State, [Stream|Tail], LastStreamID, Reason, Acc) ->
 	goaway_streams(State, Tail, LastStreamID, Reason, [Stream|Acc]).
 
+%% A server that is attempting to gracefully shut down a connection SHOULD send
+%% an initial GOAWAY frame with the last stream identifier set to 2^31-1 and a
+%% NO_ERROR code. This signals to the client that a shutdown is imminent and
+%% that initiating further requests is prohibited. After allowing time for any
+%% in-flight stream creation (at least one round-trip time), the server can send
+%% another GOAWAY frame with an updated last stream identifier. This ensures
+%% that a connection can be cleanly shut down without losing requests.
+-spec initiate_closing(#state{}, _) -> #state{}.
+initiate_closing(State=#state{http2_status=connected, socket=Socket,
+		transport=Transport, opts=Opts}, Reason) ->
+	Transport:send(Socket, cow_http2:goaway(16#7fffffff, no_error, <<>>)),
+	Timeout = maps:get(goaway_initial_timeout, Opts, 1000),
+	Message = {goaway_initial_timeout, Reason},
+	set_timeout(State#state{http2_status=closing_initiated}, Timeout, Message);
+initiate_closing(State=#state{http2_status=Status}, _Reason)
+		when Status =:= closing_initiated; Status =:= closing ->
+	%% This happens if sys:terminate/2,3 is called twice or if the supervisor
+	%% tells us to shutdown after sys:terminate/2,3 is called or vice versa.
+	State;
+initiate_closing(State, Reason) ->
+	terminate(State, {stop, stop_reason(Reason), 'The connection is going away.'}).
+
+%% Switch to 'closing' state and stop accepting new streams.
+-spec closing(#state{}, Reason :: term()) -> #state{}.
+closing(State=#state{streams=Streams}, Reason) when Streams =:= #{} ->
+	terminate(State, Reason);
+closing(State=#state{http2_status=closing_initiated,
+		http2_machine=HTTP2Machine0, socket=Socket, transport=Transport},
+		Reason) ->
+	%% Stop accepting new streams.
+	{LastStreamID, HTTP2Machine} =
+		cow_http2_machine:set_last_streamid(HTTP2Machine0),
+	Transport:send(Socket, cow_http2:goaway(LastStreamID, no_error, <<>>)),
+	closing(State#state{http2_status=closing, http2_machine=HTTP2Machine}, Reason);
+closing(State=#state{http2_status=closing, opts=Opts}, Reason) ->
+	%% If client sent GOAWAY, we may already be in 'closing' but without the
+	%% goaway complete timeout set.
+	Timeout = maps:get(goaway_complete_timeout, Opts, 3000),
+	Message = {goaway_complete_timeout, Reason},
+	set_timeout(State, Timeout, Message).
+
+stop_reason({stop, Reason, _}) -> Reason;
+stop_reason(Reason) -> Reason.
+
 -spec terminate(#state{}, _) -> no_return().
 terminate(undefined, Reason) ->
 	exit({shutdown, Reason});
 terminate(State=#state{socket=Socket, transport=Transport, http2_status=Status,
 		http2_machine=HTTP2Machine, streams=Streams, children=Children}, Reason)
-		when Status =:= connected; Status =:= closing ->
+		when Status =:= connected; Status =:= closing_initiated; Status =:= closing ->
 	%% @todo We might want to optionally send the Reason value
 	%% as debug data in the GOAWAY frame here. Perhaps more.
-	case Status of
-		connected ->
+	if
+		Status =:= connected; Status =:= closing_initiated ->
 			Transport:send(Socket, cow_http2:goaway(
 				cow_http2_machine:get_last_streamid(HTTP2Machine),
 				terminate_reason(Reason), <<>>));
 		%% We already sent the GOAWAY frame.
-		closing ->
+		Status =:= closing ->
 			ok
 	end,
 	terminate_all_streams(State, maps:to_list(Streams), Reason),
@@ -1134,9 +1206,9 @@ system_continue(_, _, {State, Buffer}) ->
 	loop(State, Buffer).
 
 -spec system_terminate(any(), _, _, {#state{}, binary()}) -> no_return().
-system_terminate(Reason, _, _, {State, _}) ->
-	%% @todo Graceful shutdown here as well?
-	terminate(State, {stop, {exit, Reason}, 'sys:terminate/2,3 was called.'}).
+system_terminate(Reason0, _, _, {State, Buffer}) ->
+	Reason = {stop, {exit, Reason0}, 'sys:terminate/2,3 was called.'},
+	loop(initiate_closing(State, Reason), Buffer).
 
 -spec system_code_change(Misc, _, _, _) -> {ok, Misc} when Misc::{#state{}, binary()}.
 system_code_change(Misc, _, _, _) ->
--- a/test/http2_SUITE.erl
+++ b/test/http2_SUITE.erl
@@ -29,6 +29,8 @@ init_dispatch(_) ->
 	cowboy_router:compile([{"localhost", [
 		{"/", hello_h, []},
 		{"/echo/:key", echo_h, []},
+		{"/delay_hello", delay_hello_h, 500},
+		{"/long_delay_hello", delay_hello_h, 5000},
 		{"/resp_iolist_body", resp_iolist_body_h, []}
 	]}]).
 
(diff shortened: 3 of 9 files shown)


essen:
Weird, I thought I implemented this. :-) Oh well, must have done it only in Gun.

> A GOAWAY frame with the last stream id set to 2^31-1 is sent and
> a timer is started, which shuts down the connection if it hasn't
> been closed already. Typically, the client will close the connection
> after receiving responses for all ongoing requests.

You should not send the max streamid as this would make the streamid value both invalid and useless. What you should do, is send a GOAWAY with the current last_streamid, and then when actually closing the connection send a second GOAWAY frame with the final last_streamid. Sending multiple GOAWAY frames is not an error.

The shutdown timer default value should be less than what the supervisor or sys:terminate call will accept to wait by default.

I have not yet looked at the code.

zuiderkwast:
No, the RFC is quite specific on this:

<blockquote>
A server that is attempting to gracefully shut down a
   connection SHOULD send an initial GOAWAY frame with the last stream
   identifier set to 2^31-1 and a NO_ERROR code.  This signals to the
   client that a shutdown is imminent and that initiating further
   requests is prohibited.  After allowing time for any in-flight stream
   creation (at least one round-trip time), the server can send another
   GOAWAY frame with an updated last stream identifier.  This ensures
   that a connection can be cleanly shut down without losing requests.
</blockquote>

essen:
Ah right.

zuiderkwast:
In master, the `shutdown_timeout` is already implemented but only for server-initiated streams, which are shutdown when the client sends goaway, for streams with stream ID > last stream ID in the goaway frame.

There is code calling `terminate_all_streams/3` and `cowboy_children:terminate/1` in `cowboy_http2:terminate/2`, but this doesn't give the streams any change to send responses to the client, since the flow is never coming back to the main receive `loop/2` after this point.

In my commits for server-initiated shutdown, I didn't use `cowboy_children:terminate/1`. I just wait for the streams to complete and if they don't, terminate will take care of them. It seems to work for simple cases.

There are a few different scenarios, depending on how clients can respond to the initial goaway. If they respond nicely (like Gun) with goaway and then disconnect, then it works fine and terminates quickly, but I think we need to consider also the case when the client doesn't respond with goaway. (They SHOULD or MAY but there is no MUST, as far as I can tell.)

I made a few test cases using Gun, but perhaps I should add some plain HTTP/2 binary communication test cases to rfc7540_SUITE or so. What do you think?

Consider this a proof-of-concept ready for review. Thanks!

essen:
Yes when the client sends GOAWAY then the server will forcefully terminate the streams that the client does not consider "good", and then leave the remaining streams running until the timeout. When the server gracefully shuts down however all streams should be considered "good" so no need to forcefully terminate them before the timeout.

About the test cases, yes it would be good to have a test where the client just stays connected so the timeout can trigger, and another test where the client immediately closes the connection.

I can't look more closely right now I'll try to remember tomorrow.

src/cowboy_http2.erl
```diff
@@ -69,6 +69,10 @@
 }.
 -export_type([opts/0]).
 
+%% Estimated time of a round-trip; the time to wait after initiating graceful
+%% shutdown before we stop accepting new streams.
+-define(shutdown_rtt, 500).
```
essen, line 74: It should be a configurable option. `shutdown_something_timeout`. I don't think rtt is descriptive enough.

src/cowboy_http2.erl
```diff
@@ -238,8 +245,12 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport,
 			setopts_active(State),
 			loop(State, Buffer);
 		%% System messages.
+		{'EXIT', Parent, shutdown} ->
+			%% Our supervisor attemts to shut us down and expects us
+			%% to terminate with reason 'shutdown' or 'normal'.
```
essen, line 250: This comment is not necessary, nor is the one for the next clause.

src/cowboy_http2.erl
```diff
@@ -252,12 +263,23 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport,
...
 		%% Messages pertaining to a stream.
 		{{Pid, StreamID}, Msg} when Pid =:= self() ->
 			loop(info(State, StreamID, Msg), Buffer);
 		%% Exit signal from children.
 		Msg = {'EXIT', Pid, _} ->
-			loop(down(State, Pid, Msg), Buffer);
+			NewState = down(State, Pid, Msg),
+			if
+				NewState#state.http2_status =:= closing, NewState#state.streams =:= #{} ->
+					terminate(NewState, {stop, normal, 'The connection is going away.'});
```
essen, line 279: Do this in the down function directly.

src/cowboy_http2.erl
```diff
@@ -938,6 +964,54 @@ goaway_streams(State, [{StreamID, #stream{state=StreamState}}|Tail], LastStreamI
 goaway_streams(State, [Stream|Tail], LastStreamID, Reason, Acc) ->
 	goaway_streams(State, Tail, LastStreamID, Reason, [Stream|Acc]).
 
+%% A server that is attempting to gracefully shut down a connection SHOULD send
+%% an initial GOAWAY frame with the last stream identifier set to 2^31-1 and a
+%% NO_ERROR code. This signals to the client that a shutdown is imminent and
+%% that initiating further requests is prohibited. After allowing time for any
+%% in-flight stream creation (at least one round-trip time), the server can send
+%% another GOAWAY frame with an updated last stream identifier. This ensures
+%% that a connection can be cleanly shut down without losing requests.
```
essen, line 973: I think you should add a `closing_initiated` state that becomes `closing` when the timeout triggers. The difference between the two would be that we don't terminate the connection if all streams finish in `closing_initiated`.

src/cowboy_http2.erl
```diff
@@ -406,6 +428,10 @@ data_frame(State0=#state{opts=Opts, flow=Flow, streams=Streams}, StreamID, IsFin
 			State0
 	end.
 
+headers_frame(State=#state{http2_status=closing}, StreamID, _, _, _, _) ->
+	%% Don't accept a new stream after sending GOAWAY with a last stream ID.
+	reset_stream(State, StreamID, {stream_error, refused_stream,
+		'New streams not accepted while going away. (RFC7540 8.1.4)'});
```
essen, line 434: I don't think you need to send the error since we already sent a GOAWAY. On the other hand it must ignore streams larger than the stream_id we sent in the GOAWAY frame, including *all* frames, not just HEADERS.

src/cowboy_http2.erl
```diff
@@ -938,6 +964,54 @@ goaway_streams(State, [{StreamID, #stream{state=StreamState}}|Tail], LastStreamI
...
+%% an initial GOAWAY frame with the last stream identifier set to 2^31-1 and a
+%% NO_ERROR code. This signals to the client that a shutdown is imminent and
+%% that initiating further requests is prohibited. After allowing time for any
+%% in-flight stream creation (at least one round-trip time), the server can send
+%% another GOAWAY frame with an updated last stream identifier. This ensures
+%% that a connection can be cleanly shut down without losing requests.
+-spec graceful_shutdown(#state{}, _) -> #state{}.
+graceful_shutdown(State=#state{socket=Socket, transport=Transport, http2_status=Status}, Reason)
+		when Status =:= connected; Status =:= closing ->
+	%% @todo Save the timer ref in the state?
```
essen, line 977: Yes. Also the `idle_timeout` should probably be canceled since it's unlikely to trigger and I don't think we want to just close in the middle of shutting down gracefully. In that case you could use the `timer` field used for `idle_timeout` for this purpose, instead of making a new record field.

src/cowboy_http2.erl
```diff
@@ -938,6 +964,54 @@ goaway_streams(State, [{StreamID, #stream{state=StreamState}}|Tail], LastStreamI
...
+	graceful_shutdown_continue(State#state{http2_status=closing}, Reason);
+graceful_shutdown_continue(State=#state{http2_status=closing, opts=Opts}, Reason) ->
+	%% Give ongoing streams some time to complete.
+	case maps:get(shutdown_timeout, Opts, 5000) of
+		infinity ->
+			State; %% No hard shutdown. All streams have to finish gracefully.
+		ShutdownTimeout ->
+			%% Shutdown hard just before the supervisor kills us.
+			Timeout = max(ShutdownTimeout - ?shutdown_rtt - 100, 0),
+			_ = erlang:start_timer(Timeout, self(), {hard_shutdown, Reason}),
```
essen, line 1011: Don't do calculations like this. You want one configurable value for the time until we stop accepting new streams during shutdown, and one configurable value for the time to shutdown itself.

src/cowboy_http2.erl
```diff
@@ -238,8 +245,12 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport,
 			setopts_active(State),
 			loop(State, Buffer);
 		%% System messages.
+		{'EXIT', Parent, shutdown} ->
+			%% Our supervisor attemts to shut us down and expects us
+			%% to terminate with reason 'shutdown' or 'normal'.
+			loop(graceful_shutdown(State, shutdown), Buffer);
```
essen, line 251: There doesn't seem to be any need to provide the Reason here, and in timeout messages, it's always `shutdown`.

essen (changes requested):
Instead of graceful/hard shutdown I think we should talk about `closing_initiated` and `closing`. There's no need to have different terms for what's roughly the same operation, only with one extra step.

I have not looked at the tests yet.

src/cowboy_http2.erl
zuiderkwast: OK! (I was also thinking about calculating a default in cowboy:start_clear and start_tls by stealing the ranch shutdown and injecting it in the protocol options to make sure we shut down in time before we're killed, but I guess that's out of the question... Fine!)

This means we can't reuse the already existing `shutdown_timeout` configuration which defaults to 5000, since we would be killed by the supervisor before the timeout fires off, right?

How about a new option `init_shutdown_timeout` (default 500) and changing the default for `shutdown_timeout` to say 4000? Or would you prefer adding another option instead of changing the default of the existing one?

src/cowboy_http2.erl
zuiderkwast: `sys:terminate/2,3` expects the process to exit with the same Reason as provided to `sys:terminate/2,3` and a supervisor expects `shutdown` in particular, but `cowboy_http2` exit Reasons always seem to be wrapped in extra tuples for debugging, so this doesn't work anyway. If there is no plan to change this (i.e. to exit with the requested reason), it's better to remove the Reason parameter here, sure.

src/cowboy_http2.erl
essen: `shutdown_timeout` is about killing child processes, not about the connection shutdown. It's not dependent on the protocol. The value it has is fine.

The `shutdown` option in Ranch is the one that's relevant here, and it defaults to 5000. So the total graceful shutdown should by default take less time. 1000+3000 for example would be plenty of time.

I think that since it's something specific to HTTP/2, maybe we should name the two new options `goaway_something_timeout`. Maybe `goaway_closing_timeout` and `goaway_close_timeout`. I don't know, naming is hard.

src/cowboy_http2.erl
essen: Right, nevermind, I missed the `sys:terminate` stuff.

src/cowboy_http2.erl
zuiderkwast: OK.

A few guards checking for `Status =:= connected` will have to be changed to `Status =:= connected; Status =:= closing_initiated` but that's fine. If `sys:terminate/2,3` is called while in `closing_initiated`, it can be ignored.

I reused the `connected` state instead of introducing a new one `closing_initiated` since there isn't really any logic which needs to know the difference, but I guess the scenario where `sys:terminate/2,3` is called twice adds enough motivation for a new state, plus it's more clear to have a different state.

src/cowboy_http2.erl
zuiderkwast: OK. `initial_shutdown_timeout` or `shutdown_initiation_timeout`?

src/cowboy_http2.erl
zuiderkwast: "closing" and "close" can cause confusion I think. How about `goaway_initial_timeout` and `goaway_close_timeout`? (The RFC refers to "an initial GOAWAY frame".)

essen: I don't think it's much better but let's go with this for now. Let's see how it looks like once it has a documentation entry explaining what it's about.

zuiderkwast: OK, I'll go with that for now.

Just a thought: For HTTP/1.1, graceful shutdown can be accomplished by sending `connection: close` after finishing the current request, with a timeout for forcing it, so maybe at least one of the options could be useful outside of HTTP/2.

essen: For HTTP/1.1 besides marking the current stream as being the last, I don't think there's any need for a timeout in the supervisor shutdown case. The supervisor can kill the process. Only the `sys:terminate` case would need a timeout.

zuiderkwast: True. And for `sys:terminate`, if it doesn't finish soon enough, you could always `exit(Conn, kill)` the process, so perhaps there is no need for a timer for HTTP/1.

src/cowboy_http2.erl
zuiderkwast: Good point. It only need to be ignored for HEADERS though, because it's the only frame from the client which can create new streams. I see now that headers_frame/6 is used for push also, so I'm moving the logic to frame/2.

essen: No you need to ignore all frames because a HEADERS frame for a new stream can be immediately followed with DATA frames for example. There's no round trip before DATA frames can be sent.

zuiderkwast: DATA and RST_STREAM frames are already ignored if their StreamID is not a key in the map `State#state.streams`. HEADERS is the only frame for which we create a new key in the state streams map. Therefore, I don't think there is any other case which needs to be handled.

essen: OK. What about WINDOW_UPDATE? Also note that there might be things in the `cow_http2_machine` code.

zuiderkwast: You're right of course. `cow_http2_machine` produces an error for frames referring no non-existent streams, which are forwarded back to the client.

I added a step before frame/2. (I also added some code which probably should belong in cowlib.)

zuiderkwast:
The last update addresses all the comments. Thanks for the comments btw! Adding a new state 'closing_initiated' makes this step more explicit and clear I think.

Regarding the names of the timers, I used the names we discussed before. I have an idea for a name of the second timer though: `goaway_complete_timeout`, since the purpose is to wait for existing streams to complete. The two-step process (and the timers)  are then the "initial" and the "complete" step.

essen:
OK for the name.

zuiderkwast:
FYI: The last update only renames `goaway_close_timeout` to `goaway_complete_timeout`.

zuiderkwast:
Added the new parameters to the manual.

zuiderkwast:
Graceful shutdown for HTTP/1.x done as discussed earlier.

src/cowboy_http2.erl
```diff
@@ -335,11 +353,42 @@ frame_rate(State0=#state{opts=Opts, frame_rate_num=Num0, frame_rate_time=Time},
 
 %% Frames received.
 
+%% Ignore frames for client-initiated streams as indicated in our GOAWAY.
```
essen, line 356: To be honest I'm starting to think that this should be handled directly in `cow_http2_machine`, not in Cowboy/Gun. So instead of only getting the LastStreamID when we initiate the close we would tell `cow_http2_machine` we are closing, and it will automatically ignore frames coming on later streams (perhaps with a limit on how much we are willing to ignore, later).

zuiderkwast: OK, then `cow_http2_machine:frame/2` would just return `{ok, MachineState}` for ignored frames, I assume.

But `cow_http2_machine` would have to know we are closing, which means we need new field in `#http2_machine{}`?

essen: I was going to say add a value to `state`, but if we do that we won't be able to catch CONTINUATION errors. We still want to detect all protocol errors even if we are closing.

So yes, I suppose a new field is necessary. Maybe we can have a `remote_last_streamid` field that initially defaults to the max value (so no stream is ignored) and is set to the `remote_streamid` value when we send the second GOAWAY frame. So technically `cow_http2_machine` won't need to know about the details of the graceful shutdown implementation, it'll just know that it must not return frames above that value.

The function updating `remote_last_streamid` can return the value at the same time so a GOAWAY can be sent without having to call the `get_` function. We should also make sure that `cow_http2_machine` crashes if the user tries to change the value to a higher number since that's not allowed, at least as far as GOAWAYs are concerned. This'll prevent errors.

zuiderkwast: Great!

If we always want to set `remote_last_streamid` to the current value of `remote_streamid`, there is no need to pass a value. Only passing the state would be enough. Then name `set_last_streamid/1` is weird if it doesn't take the value as a parameter, so maybe a better name would be `assign_last_streamid` or something like `stop_accepting_new_streams/1`?

This means `get_last_streamid/1` is obsoleted, I guess. I don't think it's used for anything else. Maybe it should be deprecated?

Another option is to add an `set_last_streamid/2` which just returns the new state (or `error(badarg)`). If you pass the value, it doesn't need to be returned, does it?

essen: I think `set_last_streamid/1` is fine because it doesn't preclude adding a second argument later on to set a custom value, for example if an implementation has good reasons to want to reduce the value even further. It also does what it says, only using a value found in the state.

I don't think `get_last_streamid/1` is obsolete, because we still need it for connection errors (Cowboy's `terminate` function for example). No need to update the HTTP2Machine state in that case. Anyway there's no need to deprecate the function because it's not documented to begin with.

zuiderkwast:
This PR currently depends on cowlib#109. When that is merged, this PR can depend on cowlib master or a new cowlib tag. Any preference?

essen:
Cowlib master but make sure to write it in the commit message.

zuiderkwast:
OK. The commit which depends on cowlib master is prepared and pushed to a separate branch: [zuiderkwast:graceful_shutdown_cowlib_master](https://github.com/zuiderkwast/cowboy/tree/graceful_shutdown_cowlib_master).

I'll push it to this PR's brach ASAP when cowlib#109 is merged.

zuiderkwast:
I see the problems in the ws_handler_SUITE. I'm looking.

essen:
It's not your fault it's because the code needs to be updated for Gun 2.0.

essen:
http2_SUITE:graceful_shutdown_connection fails for me. I'll update master to fix the issues unrelated to this PR.

zuiderkwast:
> http2_SUITE:graceful_shutdown_connection fails for me.

It doesn't fail for me.

> I'll update master to fix the issues unrelated to this PR.

That's great. I'm reading the Gun realeast notes "Migrating from 1.3" and I see that gun:ws_send/2 is replaced by gun:ws_send/3. I can try to update the ws_handler_SUITE with this if you're not already doing it.

essen:
I do it.

Does sys_SUITE fails for you?

zuiderkwast:
> I do it.

OK

> Does sys_SUITE fails for you?

No, only examples_SUITE (15 failed) and ws_handler_SUITE (16 failed). I deleted the deps directory before running tests. I'm on debian testing with OTP 23.1.2.

essen:
OK hold on because I have a bunch of failures that I'm not expecting.

essen:
OK pushed, master is green again for me.

zuiderkwast:
Great. Rebased. It's green for me.

test/http2_SUITE.erl
```diff
@@ -284,3 +286,128 @@ settings_timeout_infinity(Config) ->
...
+		<<"Hello world!">> = iolist_to_binary(RespBody),
+
+		%% Check that the connection is gone soon afterwards. (The exit
+		%% reason is supposed to be 'goaway' as passed to
+		%% sys:terminate/2, but it is {shutdown, {socket_error, closed},
+		%% 'The socket has been closed.'}}.)
+		receive
+			{'DOWN', _, process, CowboyConnPid, _Reason} ->
+				ok
+		after 100 ->
```
essen, line 322: Remove the after clause. Timeouts lead to intermittent test failures. If there's a problem the timetrap timeout will catch it.

test/http2_SUITE.erl
```diff
@@ -284,3 +286,128 @@ settings_timeout_infinity(Config) ->
...
+		{error, {stream_error, closed}} = gun:await(ConnPid, Ref),
+
+		%% Check that the connection is gone. (The exit reason is
+		%% supposed to be 'goaway' as passed to sys:terminate/2, but it
+		%% is {shutdown, {stop, {exit, goaway}, 'Graceful shutdown timed
+		%% out.'}}.)
+		receive
+			{'DOWN', _, process, CowboyConnPid, _Reason} ->
+				ok
+		after 100 ->
```
essen, line 362: Remove the after clause.

test/http2_SUITE.erl
```diff
@@ -284,3 +286,128 @@ settings_timeout_infinity(Config) ->
...
+		goaway_complete_timeout => 500
+	},
+	{ok, _} = cowboy:start_clear(?FUNCTION_NAME, [{port, 0}], ProtoOpts),
+	Port = ranch:get_port(?FUNCTION_NAME),
+	try
+		ConnPid = gun_open([{type, tcp}, {protocol, http2}, {port, Port}|Config]),
+		Ref = gun:get(ConnPid, "/long_delay_hello"),
+
+		%% Make sure the request is received.
+		timer:sleep(100),
```
essen, line 345: You should have Cowboy send the test process a message when it has received the request. Timeouts should be avoided.

test/http2_SUITE.erl
```diff
@@ -284,3 +286,128 @@ settings_timeout_infinity(Config) ->
...
+	after
+		cowboy:stop_listener(?FUNCTION_NAME)
+	end.
+
+graceful_shutdown_timeout(Config) ->
+	doc("Check that a connection is closed when gracefully shutting down times out."),
+	ProtoOpts = #{
+		env => #{dispatch => init_dispatch(Config)},
+		goaway_initial_timeout => 200,
+		goaway_complete_timeout => 500
```
essen, line 336: Why 200 and 500?

test/http2_SUITE.erl
```diff
@@ -284,3 +286,128 @@ settings_timeout_infinity(Config) ->
...
+		env => #{dispatch => init_dispatch(Config)}
+	},
+	{ok, _} = cowboy:start_clear(?FUNCTION_NAME, [{port, 0}], ProtoOpts),
+	Port = ranch:get_port(?FUNCTION_NAME),
+
+	ConnPid = gun_open([{type, tcp}, {protocol, http2}, {port, Port}|Config]),
+	Ref = gun:get(ConnPid, "/delay_hello"),
+
+	%% Shutdown listener while the handlers are working.
+	timer:sleep(100),
```
essen, line 383: Avoid the timeout.

test/http2_SUITE.erl
```diff
@@ -284,3 +286,128 @@ settings_timeout_infinity(Config) ->
...
+		goaway_complete_timeout => 500
+	},
+	{ok, _} = cowboy:start_clear(?FUNCTION_NAME, [{port, 0}], ProtoOpts),
+	Port = ranch:get_port(?FUNCTION_NAME),
+
+	ConnPid = gun_open([{type, tcp}, {protocol, http2}, {port, Port}|Config]),
+	Ref = gun:get(ConnPid, "/long_delay_hello"),
+
+	%% Shutdown listener while the handlers are working.
+	timer:sleep(100),
```
essen, line 407: Avoid the timeout.

test/http2_SUITE.erl
```diff
@@ -284,3 +286,128 @@ settings_timeout_infinity(Config) ->
...
+	ok = cowboy:stop_listener(?FUNCTION_NAME),
+
+	%% Check that the request is handled before shutting down.
+	{response, nofin, 200, _RespHeaders} = gun:await(ConnPid, Ref),
+	{ok, RespBody} = gun:await_body(ConnPid, Ref),
+	<<"Hello world!">> = iolist_to_binary(RespBody),
+
+	gun:close(ConnPid).
```
essen, line 391: You should check that the listener was actually stopped. Same for the following test.

test/http_SUITE.erl
```diff
@@ -443,3 +446,71 @@ switch_protocol_flush(Config) ->
...
+	    "shutting down a connection."),
+		ProtoOpts = #{
+		env => #{dispatch => init_dispatch(Config)}
+	},
+	{ok, _} = cowboy:start_clear(?FUNCTION_NAME, [{port, 0}], ProtoOpts),
+	Port = ranch:get_port(?FUNCTION_NAME),
+	try
+		ConnPid = gun_open([{type, tcp}, {protocol, http}, {port, Port}|Config]),
+		{ok, http} = gun:await_up(ConnPid),
+		timer:sleep(500),
```
essen, line 461: Nope.

test/http_SUITE.erl
```diff
@@ -443,3 +446,71 @@ switch_protocol_flush(Config) ->
...
+		#{socket := Socket} = gun:info(ConnPid),
+		CowboyConnPid = get_remote_pid_tcp(Socket),
+		CowboyConnRef = erlang:monitor(process, CowboyConnPid),
+		Ref1 = gun:get(ConnPid, "/delay_hello"),
+		Ref2 = gun:get(ConnPid, "/delay_hello"),
+
+		timer:sleep(100),
```
essen, line 468: Nope.

test/http_SUITE.erl
```diff
@@ -443,3 +446,71 @@ switch_protocol_flush(Config) ->
...
+		<<"close">> = proplists:get_value(<<"connection">>, RespHeaders),
+		{ok, RespBody} = gun:await_body(ConnPid, Ref1),
+		<<"Hello world!">> = iolist_to_binary(RespBody),
+
+		{error, {stream_error, _}} = gun:await(ConnPid, Ref2),
+		ok = gun_down(ConnPid),
+		receive
+			{'DOWN', CowboyConnRef, process, CowboyConnPid, _Reason} ->
+				ok
+		after 1000 ->
```
essen, line 481: Nope.

test/http_SUITE.erl
```diff
@@ -443,3 +446,71 @@ switch_protocol_flush(Config) ->
...
+	Port = ranch:get_port(?FUNCTION_NAME),
+
+	ConnPid1 = gun_open([{type, tcp}, {protocol, http}, {port, Port}|Config]),
+	Ref1 = gun:get(ConnPid1, "/delay_hello"),
+
+	ConnPid2 = gun_open([{type, tcp}, {protocol, http}, {port, Port}|Config]),
+	Ref2 = gun:get(ConnPid2, "/long_delay_hello"),
+
+	%% Shutdown listener while the handlers are working.
+	timer:sleep(100),
```
essen, line 504: Nope.

test/rfc7540_SUITE.erl
```diff
@@ -2955,11 +2957,10 @@ client_settings_disable_push(Config) ->
 %% (RFC7540 6.8) GOAWAY
 % @todo GOAWAY frames have a reserved bit in the payload that must be ignored.
 %
-%% @todo We should eventually implement the mechanism for gracefully
+%% The following test cases test the mechanism for gracefully
 %% shutting down of the connection. (Send the GOAWAY, finish processing
 %% the current set of streams, give up after a certain timeout.)
```
essen, line 2962: Just remove the todo entirely.

test/rfc7540_SUITE.erl
```diff
@@ -2975,8 +2976,10 @@ client_settings_disable_push(Config) ->
 %   GOAWAY frame with an updated last stream identifier.  This ensures
 %   that a connection can be cleanly shut down without losing requests.
 %
-%% @todo And of course even if we shutdown we need to be careful about
-%% the connection state.
+%% And of course even if we shutdown we need to be careful about the
+%% connection state. In state 'closing', we don't accept new streams
+%% because we have already sent a GOAWAY frame with the last stream ID.
```
essen, line 2981: Just remove the todo. Remove the todo comments in general for things that have been implemented and tested in this test suite.

test/rfc7540_SUITE.erl
```diff
@@ -2988,6 +2991,83 @@ client_settings_disable_push(Config) ->
 %   cause flow control or header compression state to become
 %   unsynchronized.
 %
+
+alpn_graceful_shutdown_client_stays(Config) ->
```
essen, line 2995: Test has nothing to do with `alpn_` so remove this part.

test/rfc7540_SUITE.erl
```diff
@@ -2988,6 +2991,83 @@ client_settings_disable_push(Config) ->
...
+	{ok, << 0:24, 4:8, 1:8, 0:32 >>} = ssl:recv(Socket, 9, 1000),
+	%% Server-side application logic decides to gracefully shutdown the
+	%% connection (for whatever reason).
+	ServerConnPid = get_remote_pid_tls(Socket),
+	ok = sys:terminate(ServerConnPid, whatever),
+	%% Expect a GOAWAY with last stream id 0x7FFFFFF and NO_ERROR
+	{ok, << _:24, 7:8, 0:8, 0:1, 0:31, %% Length, type, flags, R, stream id
+		0:1, 16#7fffffff:31, 0:32  %% R, last stream id, error code
+		>>} = ssl:recv(Socket, 17, 500),
+	%% The client doesn't respond. The server waits a bit (1 RTT) before
```
essen, line 3017: Not 1 RTT.

test/rfc7540_SUITE.erl
```diff
@@ -2988,6 +2991,83 @@ client_settings_disable_push(Config) ->
...
+	%% sending a 2nd GOAWAY frame with an actual last stream id = 0
+	{ok, << _:24, 7:8, 0:8, 0:1, 0:31, %% Length, type, flags, R, stream id
+		0:1, 0:31, 0:32            %% R, last stream id, error code
+		>>} = ssl:recv(Socket, 17, 1500),
+	{error, closed} = ssl:recv(Socket, 3, 1000),
+	ok.
+
+alpn_graceful_shutdown_race_condition(Config) ->
+	doc("Graceful shutdown where the client sends requests while the server "
+		"sends goaway. (RFC7540 6.8)"),
```
essen, line 3027: GOAWAY

essen (changes requested):
Let's merge both Cowlib PR and this at the same time. There's a few minor things left to fix. Did a full review so most of everything should have been catched now.

test/http2_SUITE.erl
zuiderkwast: It makes the test complete faster than with the default 1000 and 3000. Do you prefer using the defaults?

essen: No just wondering.

test/rfc7540_SUITE.erl
zuiderkwast: The alpn prefix makes it run with TLS because of logic in groups/0. I thought it's more realistic to test it with TLS, but can remove it.

essen: This test suite only uses TLS where relevant with regard to the RFC. Other test suites use TLS for just about everything you can do with Cowboy.

src/cowboy_http2.erl
```diff
@@ -938,21 +964,62 @@ goaway_streams(State, [{StreamID, #stream{state=StreamState}}|Tail], LastStreamI
...
+-spec initiate_closing(#state{}, _) -> #state{}.
+initiate_closing(State=#state{http2_status=connected, socket=Socket,
+		transport=Transport, opts=Opts}, Reason) ->
+	Transport:send(Socket, cow_http2:goaway(16#7FFFFFFF, no_error, <<>>)),
+	Timeout = maps:get(goaway_initial_timeout, Opts, 1000),
+	Message = {goaway_initial_timeout, Reason},
+	set_timeout(State#state{http2_status=closing_initiated}, Timeout, Message);
+initiate_closing(State=#state{http2_status=Status}, _Reason)
+		when Status =:= closing_initiated; Status =:= closing ->
+	%% This happens e.g. if sys:terminate/2,3 is called twice.
```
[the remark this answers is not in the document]
zuiderkwast: It can also happen if the supervisor call `exit(Conn, shutdown)` after `sys:terminate/2,3` is called. But it's a corner case and I suppose it's no problem if we get abrupt shutdown in this case. I'll move this status check to system_terminate.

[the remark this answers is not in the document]
essen: No it's fine then just update the comment to include supervisor shutdown.

src/cowboy_http2.erl
```diff
@@ -227,8 +229,11 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport,
 	receive
 		%% Socket messages.
 		{OK, Socket, Data} when OK =:= element(1, Messages) ->
-			parse(set_timeout(State), << Buffer/binary, Data/binary >>);
+			parse(set_idle_timeout(State), << Buffer/binary, Data/binary >>);
 		{Closed, Socket} when Closed =:= element(2, Messages) ->
+			%% @todo This is not an error if the client closes the
+			%% socket when Status is closing (after GOAWAY), so we
+			%% can exit with reason 'normal' in this case.
```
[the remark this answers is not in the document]
zuiderkwast: I'll add a case on `http2_status` here, but `terminate/2` wraps the reason in a shutdown tuple as `exit({shutdown, Reason})`, so we don't exit with the specified Reason anyway.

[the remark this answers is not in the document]
essen: `{shutdown,...}` is a normal exit reason as far as OTP is concerned. `{shutdown, closed}` would be fine.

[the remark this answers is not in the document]
zuiderkwast: That's true in the sense that no error is logged. Then we don't need to do anything here because terminate always exits with a shutdown tuple, but I can do `{shutdown, closed}` for debugging reasons.

In some situations, a process is expected to terminate with an exact reason, such when a supervisor tells a process to shutdown here: https://github.com/erlang/otp/blob/master/lib/stdlib/src/supervisor.erl#L873 or when [sys:terminate/2,3](https://erlang.org/doc/man/sys.html#terminate-2) is called with a specified Reason, that exact reason is supposed to be the exit reason.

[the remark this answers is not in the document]
essen: The comment says it's not an error *if we are already closing* and I agree. So the reason should mirror server-initiated closes, perhaps with a slight twist from `normal` to `closed` to tell us it's the client that closed and not us. So `{stop, closed, 'The client is going away.'}` or something to that effect, when we are already closing.

When we are not closing however the `socket_error` reason makes more sense because it's an unexpected close.

So depending on the `http2_status` we want one reason or the other.

I do not understand your comment about an exact reason, we are talking about the socket being closed here, there's no constraints.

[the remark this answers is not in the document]
zuiderkwast: I'll do as you suggested.

Just to explain the previous comment: There are no constraints in the socket closed case, but the comment was generally about wrapping any reason in `{shutdown, ...}`. For unexpected errors, shouldn't an error be logged by OTP, if the process is supposed to behave like OTP processes do? (I'm not saying an unexpected socket close is an unexpected error; it's quite normal.)

Since ranch is used and not the OTP supervisor, the point about supervisor shutdown isn't valid. The point about `sys:terminate/2,3` is, but it's a rather unimportant point. If the caller calls `sys:terminate/2,3` with an abnormal reason, I guess an error is expected to be logged.

test/http2_SUITE.erl
zuiderkwast: OK, I'll do it by monitoring the listener and receiving its 'DOWN' message.

test/http2_SUITE.erl
zuiderkwast: I'll add this notification functionality in `delay_hello_h`.

test/rfc7540_SUITE.erl
zuiderkwast: Where does this todo actually end? There is some quoted RFC text with another todo `@todo And it get's more complex when you think about h1 to h2 proxies.` and then some more quoted RFC text which is now implemented. Then, there is some text which is not marked as a todo, but it wasn't implemented before this PR. Should the comments about everything that is implemented be deleted?

zuiderkwast: I'm keeping the quoted RFC text.

zuiderkwast:
I've addressed the comments. This PR is still based on my cowlib PR branch. (The only differences are the commit message, the cowlib line in Makefile and the generated rebar.config.)

src/cowboy_http2.erl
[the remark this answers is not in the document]
essen: Ranch can shut down via `shutdown` just like the OTP supervisor.

When we crash there are appropriate error events. When the user code crash we produce easier to read error reports.

test/rfc7540_SUITE.erl
essen: The relevant documentation should be given to the `doc/1` function. It's OK if it takes 3 lines if it's something complex. Not much point keeping it in comments hidden in the middle of the test file, best to have it in log reports.

These todo comments are for things left to implement (or test), so whatever has a test shouldn't be a comment anymore. I'll double check them when merging.

src/cowboy_http2.erl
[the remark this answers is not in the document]
zuiderkwast: Yeah, the difference is that `ranch_conns_sup` doesn't care about the Reason in the `{'DOWN', _, process, Pid, _}` message while `supervisor` requires the Reason to be the atom `shutdown` or otherwise it's regarded it as an error which is logged. (Now, I don't think anyone wants to use the regular OTP `supervisor` with a Cowboy connection process.)

I'm sure the error handling/reporting/logging is appropriate as it is. Cowboy's exit reasons are surely more readable when debugging.

src/cowboy_http.erl
```diff
@@ -246,7 +246,7 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport, opts=Opts,
 			loop(State);
 		%% System messages.
 		{'EXIT', Parent, Reason} ->
-			terminate(State, {stop, {exit, Reason}, 'Parent process terminated.'});
+			initiate_closing(State, Reason);
```
essen, line 249: Please restore the stop tuple so the exit reason is the same for HTTP/1.1 and HTTP/2.

src/cowboy_http.erl
```diff
@@ -1503,9 +1510,9 @@ terminate_linger_loop(State=#state{socket=Socket}, TimerRef, Messages) ->
 system_continue(_, _, State) ->
 	loop(State).
 
--spec system_terminate(any(), _, _, {#state{}, binary()}) -> no_return().
+-spec system_terminate(any(), _, _, #state{}) -> no_return().
 system_terminate(Reason, _, _, State) ->
-	terminate(State, {stop, {exit, Reason}, 'sys:terminate/2,3 was called.'}).
+	initiate_closing(State, Reason).
```
essen, line 1515: Same here.

src/cowboy_http2.erl
```diff
@@ -238,8 +244,9 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport,
 			setopts_active(State),
 			loop(State, Buffer);
 		%% System messages.
+		{'EXIT', Parent, shutdown} ->
+			loop(initiate_closing(State, shutdown), Buffer);
```
essen, line 248: With the above comments I suspect we'll need to add this clause to HTTP/1.1 as well. By the way we might also need to do this with `cowboy_websocket`. Perhaps this new shutdown clause should be added everywhere in a separate commit that explains why we are doing this, instead of together with the other changes.

src/cowboy_http2.erl
```diff
@@ -568,7 +586,7 @@ timeout(State=#state{http2_machine=HTTP2Machine0}, Name, TRef) ->
 %% Erlang messages.
 
 down(State=#state{opts=Opts, children=Children0}, Pid, Msg) ->
-	case cowboy_children:down(Children0, Pid) of
+	NewState = case cowboy_children:down(Children0, Pid) of
```
essen, line 589: Nitpick: State0 and State, rather than State and NewState.

essen (changes requested):
A few more comments. Should be good after that.

src/cowboy_http.erl
zuiderkwast: OK, will do.

Btw, for HTTP/2 I added a separate clause for `shutdown` but not here. Should I? I guess we don't want graceful shutdown if the parent exits with other reasons, e.g. `killed`.

src/cowboy_http2.erl
zuiderkwast: So you agree we want graceful shutdown for `shutdown` and not for other Parent exit reasons?

I think graceful shutdown for Parent shutdown is the main trigger of the Graceful shutdown feature, so it would be a pity to postpone that. I can add it for websocket as well if it's strait-forward.

zuiderkwast: Btw, the exit reason after successful graceful shutdown will be `{shutdown, shutdown}` in this case, since terminate/2 wraps it like this. We could as well shutdown with a verbose tuple `{stop, {exit, shutdown}, 'Shutdown requested by parent process.'}` similar to the other cases. It's up to you.

In `cowboy_websocket`, the shutdown reasons are not verbose stop tuples like this.

essen: Just make `terminate` use `shutdown` when `Reason` is `shutdown`, and `{shutdown, Reason}` otherwise.

I don't think there's anything to postpone, I'm talking about a separate commit as part of this PR. Just to better show that these changes are changing more than just adding graceful shutdown. But both commits are part of the graceful shutdown feature. Alternatively be very explicit about this in the commit message.

src/cowboy_http.erl
essen: I'm OK with no graceful shutdown for other reasons for the time being.

src/cowboy_http2.erl
zuiderkwast: One more thing. This Reason passed to `initiate_closing/2` is only stored in timeout messages, not in the state. A graceful shutdown can end up terminating in `parse/2`, where we don't have the requested reason:

```Erlang
		more when Status =:= closing, Streams =:= #{} ->
			terminate(State, {stop, normal, 'The connection is going away.'});
```

We could either terminate with reason `normal` (which is also accepted by supervisor) or store the requested reason in the state to be able to use it here. If we want to consider the Reason passed to `sys:terminate/2,3`, we would have to do the latter.

We can also terminate in `down/3` (exit signal from children), which may also be a graceful shutdown:

```Erlang
	if
		State#state.http2_status =:= closing, State#state.streams =:= #{} ->
			terminate(State, {stop, normal, 'The connection is going away.'});
```

... or we can leave it as it is and forget about requested reason. What do you prefer?

essen: Leave those two cases as-is for now and let's see if that bothers anyone.

zuiderkwast: OK. Btw, when I added a case for `shutdown` in `terminate/2`, Dialyzer tells me that `terminate(_, shutdown)` can never be called. It turns out it is always wrapped again in a stop tuple (e.g. `{stop, Reason, 'Graceful shutdown timed out.'}`) or replaced by another reason. Most often it seems (at least with Gun) that when shutdown is initiated, the client responds with GOAWAY and then closes the connection when all streams are done, so the Reason in terminate/2 will be `{stop, closed, 'The client is going away.'}`.

Because of this, I'm leaning towards skipping the shutdown with reason `shutdown`. If we're to write in a specific commit message that we shutdown with reason `shutdown` when requested by the supervisor, we should actually do that. Otherwise, there's no point. For now, I think it's better to stick to a stop tuple:

		{'EXIT', Parent, shutdown} ->
			Reason = {stop, {exit, shutdown}, 'Parent process requested shutdown.'}
			loop(initiate_closing(State, Reason), Buffer);

and then later, if it's important, rework this by storing the requested exit reason in the state.

essen: OK.

zuiderkwast:
I did as we said. In some cases, the exit reason can now be a nested stop tuple like `{stop, {stop, {exit, shutdown}, 'Parent process requested shutdown.'}, 'Graceful shutdown timed out.'}`, which is fine I suppose.

The build fails now because it depends on Cowlib master now. I made it like this in the last update, to make it ready to submit together with the Cowlib PR.

essen:
Shouldn't have nested stop tuples. Take the `Reason` (second element) from the nested tuple, not the entire tuple.

zuiderkwast:
Done.

essen:
Can you set Cowlib back to the branch so that it can be tested with the most recent changes? I'll just switch it back to master during merge.

zuiderkwast:
Sure, done. (The commit message mention of cowlib master is currently a lie then.)

essen:
Does `make ct-http t=clear:graceful_shutdown_listener` reliably work for you?

essen:
Merged, thanks!

I edited the tests a bit. Please review them especially the newly added todos. It would be a good idea to strengthen them in the future.

zuiderkwast:
Nice! Thanks for your reviews and patience! The result is quite good I think. :tada:

Yes, `make ct-http t=clear:graceful_shutdown_listener` works reliably for me.

I checked the diff compared to my branch. Replacing sleep in some test cases is good; I didn't realize I did this only in some places. (I guess I only changed where you commented.)

I agree with the added todos. It would also be good to have unit tests for the connection state in Cowlib (cow_http2_machine).

essen:
Unit tests in cow_http2_machine is a waste of time because it is covered by both Cowboy and Gun. If I had infinite time I would perhaps do it, but I'd rather focus on the rfc7540_SUITE(s) in both projects.

PR #1477, opened 2020-11-02
Use defines for default values

ampleyfly:
... to ensure that the same values are used in all places.
This was not the case for one use of max_received_frame_rate.


essen:
Please fix the value if you think it's wrong but do not use defines.

I can see another problem which is that the code does not use `add_period/2` everywhere.

Perhaps have functions `init_max_received_frame_rate` and `init_max_reset_stream_rate` that receive State and CurrentTime and return an updated state.

ampleyfly:
I did an update with something similar to what you proposed. There seems to be something wrong with the test environment, though :(

essen:
I'll fix it.

essen:
Merged, thanks!

PR #1480, opened 2020-11-04
Don't hang forever if stream proc is down

ampleyfly:
If the stream process---the 'pid' in the cowboy request---has
terminated, don't hang forever waiting for an ack.

The stream process could presumably terminate if the remote end
closes the tcp connection.

I think the test is a bit iffy, so if there's a better way to do it, suggestions are welcome.

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -887,7 +887,14 @@ stream_body(Data, IsFin, Req=#{has_sent_resp := headers})
 %% @todo Do we need a timeout?
 stream_body(Msg, Req=#{pid := Pid}) ->
 	cast(Msg, Req),
-	receive {data_ack, Pid} -> ok end.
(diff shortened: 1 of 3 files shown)


essen:
I don't think that's right. If the connection is down, that means the connection pid terminates, and therefore the request pid will terminate as well, unless you did something weird with links or there's a bug of course. Behavior may differ when trapping exits of course and there might be improvements to be done there.

ampleyfly:
Yes, this becomes a problem if the request pid is handed to some other process which deals with streaming the body, and the connection dies.

essen:
OK you meant using Req from a different process. I don't think this is supposed to work at the moment. There's probably many more issues.

ampleyfly:
Ok, I will close it. We will have to fix the use on our side, instead.

Issue #1430, opened 2019-12-25
cowboy_stream_h problems in grpc streaming


tony612:
I got two problems when using cowboy_stream_h to implement grpc streaming calls:
1. I tried to use `auto` mode, but got an error when calling `cowboy_req:read_body(Req, [{length, auto}])`. I believe this is because this line https://github.com/ninenines/cowboy/blob/13cf0baa41ad0b6e96f7fea0b95a8002e930e4a2/src/cowboy_req.erl#L520
2. Streaming client may send only one empty body with `fin` to info indicate the streaming request is done. But current `cowboy_stream_h` can't handle this because https://github.com/ninenines/cowboy/blob/13cf0baa41ad0b6e96f7fea0b95a8002e930e4a2/src/cowboy_stream_h.erl#L161-L166 To solve this, we may need to send the body back when `fin`.

essen:
You can't use read_body with auto at the moment (or ever, to be honest, it's not really the same thing) but you can send the message directly. Now there's a `cowboy_req:cast(Req, Msg)` to send that message.

It should special case `fin` indeed.

essen:
I am thinking of reworking `read_body auto` into a sort of `active,N` where we would be able to send a message asking for the body to be sent as it comes in for a total of N messages, followed by a type of passive message like `active,N` has. Then it can be requested again. It could also be canceled if necessary.

I think we should also be able to request a minimum flow value when doing this. So perhaps a command like `min_flow` needs to be added. Then if we set `min_flow` and there's already a high enough flow value there's no need to send `WINDOW_UPDATE` or other and no risk to overflow.

I am thinking of making new messages rather than reuse `read_body`. Maybe `{active_read_body, Pid, Ref, MinFlow, N}` where `N` can be `false` to cancel. The `request_body` message can be kept as is, however an extra message `{passive_read_body, Ref}` needs to be sent once `N` reaches 0 or is otherwise canceled. Sending `active_read_body` twice in a row would result in the `N` values being added up and the `MinFlow` updated.

tony612:
@essen Then we need to store the data as a list of binary instead binary like what we do now, right? And when `read_body` is called, we should call `iolist_to_binary` first?

essen:
In active mode, you don't call `read_body`.

And either way there's never a need to use lists of binaries when you can just do binary appends. That's how `read_body` works.

PR #1482, opened 2020-11-05
CONNECT method (Proof-of-Concept)

zuiderkwast:
I've been experimenting with CONNECT to write a proxy server: https://github.com/zuiderkwast/deputy. (Is the name Deputy already taken or reserved for something else in the Cowboy world?) I've tested it only manually by configuring it as a proxy running on localhost in Firefox.

**Feel free to pick selected stuff from this PR, now or later. I don't really need it for any serious project.**

Changes to Cowboy:

* Request parsing and validation for CONNECT update in `cowboy_http` and `cowboy_http2` and restrictions forbidding CONNECT removed. Path is set to "/" rather than omitted, to make the request pass through `cowboy_router` and `cowboy_handler` unmodified. (It's a hack.)
* A `cowboy_tunnel` modelled after `cowboy_websocket` is added.
* Adaptations to `cowboy_http` for taking over the connection process without sending *101 Switching Protocols*.

The flow, in short:

1. `Handler:init` is called. For CONNECT requests, it may return `{cowboy_tunnel, ...}`.
2. `cowboy_tunnel:upgrade` is called. It calls cowboy_http with a new command `{takeover, ...}` which has the same effect as `{switch_protocol, ...}`, except 101 is not sent. Takover is called in the connection process.
3. `cowboy_tunnel` calls `Handler:tunnel_init` in the connection process. Here, a TCP connection (tunnel) is supposed to be created.
4. If successful, `cowboy_tunnel` sends 200 OK to the client and the tunnel is open.
5. On info (such as incoming TCP tunnel data from the peer) `Handler:tunnel_info` is called. On data from the client, `Handler:tunnel_handle` is called.

TODO:

* Decide how the interface should actually look like
* Test cases
* Remove commented-out code

### DIFF
--- a/ebin/cowboy.app
+++ b/ebin/cowboy.app
@@ -1,7 +1,7 @@
 {application, 'cowboy', [
 	{description, "Small, fast, modern HTTP server."},
 	{vsn, "2.8.0"},
-	{modules, ['cowboy','cowboy_app','cowboy_bstr','cowboy_children','cowboy_clear','cowboy_clock','cowboy_compress_h','cowboy_constraints','cowboy_handler','cowboy_http','cowboy_http2','cowboy_loop','cowboy_metrics_h','cowboy_middleware','cowboy_req','cowboy_rest','cowboy_router','cowboy_static','cowboy_stream','cowboy_stream_h','cowboy_sub_protocol','cowboy_sup','cowboy_tls','cowboy_tracer_h','cowboy_websocket']},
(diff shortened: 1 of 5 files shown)


essen:
Just answering the comments you made, didn't look at the code yet:

* Yes `deputy` is fair game.
* The path hack is definitely not good. I was thinking a CONNECT handler should be configured in the protocol options, question is where exactly (in `env` probably, in which case `cowboy_router` can just take it from there or abort if none is configured).
* Perhaps `switch_protocol` can be replaced by `takeover` entirely.
* I'm not convinced a special handler is needed otherwise. `cowboy_loop` should be enough once initialization is done. And it can use "read_body auto" to receive packets from the socket as they come (or better https://github.com/ninenines/cowboy/issues/1430#issuecomment-571112763 but the code doesn't currently exist).

PR #1485, opened 2020-11-20
Also include trace messages when timestamp flag isn't used

sstrollo:
When I set my own custom tracer_flags I didn't use timestamp (or monotonic_timestamp), but I didn't get any trace callbacks but I did noticed a lot of "Tracer process received stray message". I think this patch fixes the issue (since without a timestamp flag the traces are sent with the atom 'trace', instead of 'trace_ts').

Thanks!

/Sebastian


essen:
Right thanks I'll merge it soon.

essen:
Merged, thanks!

PR #1494, opened 2021-01-07
Fix wrong template name

anacrolix:


essen:
It's not wrong, did you set `DEP_PLUGINS = cowboy`?

anacrolix:
I did have it set. I think I didn't have `dep_cowboy_commit = 2.8.0 set`, after putting that in and cleaning the deps, the missing template appeared. So unless the template name is changed in master or future versions, this PR should be closed, sorry.

essen:
It changed when it was moved from Erlang.mk to Cowboy itself when dep plugins were introduced. There's an Erlang.mk cleanup coming up that will remove any confusion about that soon.

anacrolix:
Thanks @essen , sorry for the noise.

Issue #1502, opened 2021-02-10
Allowing download time to not be considered as "idle"


rjmac:
Currently, the time required to send a response is counted as idle, and as a result if it takes more than one minute (by default) the idle timeout kills the request.  We've got a use case where the response is generated, and can potentially take multiple minutes to download (streamed, so it's continually producing data) and there isn't actually a fixed upper bound on the amount of time the response will take to send.

There'll be a pull request with our approach to allowing this shortly; it preserves the existing behavior but adds a new protocol option to the http and http2 backends which makes the idle timer get reset on sends.

essen:
You can set the `idle_timeout` option on a per-request basis. You have an option here: https://ninenines.eu/docs/en/cowboy/2.9/manual/cowboy_req.cast/#_examples

I haven't had time to consider the PR yet. My focus in the coming week or two will be Gun.

qhartman:
Does the `idle_timeout` logic that is applied per-request behave differently?

The issue here is that the global `idle_timeout` isn't correctly tracking ongoing network activity, so a large download that is moving data but has no additional requests associated with it is incorrectly counted as idle and gets incorrectly closed.

essen:
PR was merged, closing, thanks! New option is `reset_idle_timeout_on_send` and must be set to `true` to apply. Default behavior will probably be changed in the next major release because it only makes sense when paired with other improvements made recently.

PR #1503, opened 2021-02-10
Optionally reset the idle timeout when sending a chunk

rjmac:
* The default behavior (sending data is considered being idle) is
  maintained.
* There is a new protocol option for http and http2 which makes
  the idle timeout be reset when data is sent to the client.

This implements #1502.


essen:

essen:
I will keep the tests and do a different implementation that updates the idle timeout on all cases of sending on the socket, not just when data is sent.

essen:
Merged, thanks! Note that the option was renamed to `reset_idle_timeout_on_send`.

Issue #1511, opened 2021-04-21
Handle request processes exits that get sent exit signal


hubertlepicki:
When a web request is being processes by Cowboy, and the request handling process get sent exit signal (as in with `exit/2`), it seems like the process is silently crashing without reporting anything to the logger.

It looks like Cowboy handles exits of these processes here using try/catch clause:

https://github.com/ninenines/cowboy/blob/master/src/cowboy_stream_h.erl#L289-L300

But it does only work if process exits on it's own. When it is being sent exit signal, it's not.

This is because the process is not trapping exits, and even that wouldn't catch all signals, i.e. `kill` wouldn't get detected.

I have to detect these exits, and I am currently doing that by setting up a monitor.

I am just wondering if that is not something Cowboy would want to be doing internally? Since it catches exits of these processes if they exit with `exit/1`, maybe it makes sense to add functionality to catch abnormal exits when these processes are sent a signal with `exit/2` too?

essen:
What you linked to is only to add the stacktrace to exit reasons. The actual handling is at https://github.com/ninenines/cowboy/blob/master/src/cowboy_stream_h.erl#L127-L156

You are correct that it's missing a more general clause.

hubertlepicki:
Right, so it's probably missing the clause of 

```
info(StreamID, Exit={'EXIT', Pid, Reason})
```

Want me to give it a shot with a PR?

essen:
Sure.

hubertlepicki:
I have created a PR. There's at least one issue with it (silencing crashes in tests that I don't know how to do properly). Other than that it seems to resolve the issue at least on my app.

essen:
Done for Cowboy 2.13. Closing, thanks!!

hubertlepicki:
Awesome @essen this is still relevant to me so I'll pass that on to the team that's currently monitoring this with custom implementation. Thank you so much!

PR #1514, opened 2021-04-23
[#1551] Handle external exits of processes

hubertlepicki:
Relates to #1511 

Adds handler to capture external exits of processes, and tests to verify
this is indeed happening.

On my Elixir/Phoenix project, where I add such line in a controller to kill the process:

```
Process.exit(self(), :kill) 
```

Without this change, no exception or error is reported to the logger.

With these changes I see the error line as I'd expect:

```
14:07:06.341 [error] Ranch listener UI.Endpoint.HTTP, connection process #PID<0.14777.0>, stream 9 had its request process #PID<0.14896.0> exit with reason :killed
```

There is at least one one problem: I don't know how to silence crash reports in this case. When I run the tests, I can see the crash reports I added, but I would like to silence them in tests. ct_helper:ignore() doesn't seem to cut it the way I use it, which is probably wrong. I see on the console:

```
Testing ninenines.cowboy.metrics_SUITE: Starting test, 96 test cases

=ERROR REPORT==== 23-Apr-2021::11:22:09 ===
Ranch listener http, connection process <0.266.0>, stream 1 had its
request process <0.275.0> exit with reason external_exit
```

which we probably want to silence but I'm not sure how, so a suggestion how and a general review of my code would be very much appreciated. Unfortunately I don't write Erlang daily so I appreciate feedback.

### DIFF
--- a/src/cowboy_stream_h.erl
+++ b/src/cowboy_stream_h.erl
@@ -154,6 +154,22 @@ info(StreamID, Exit={'EXIT', Pid, {Reason, Stacktrace}}, State=#state{ref=Ref, p
 	do_info(StreamID, Exit, [
 		{error_response, 500, #{<<"content-length">> => <<"0">>}, <<>>}
 	|Commands], State);
+info(StreamID, Exit={'EXIT', Pid, Reason}, State=#state{ref=Ref, pid=Pid}) ->
+	Commands0 = [{internal_error, Exit, 'Stream process crashed.'}],
+	Commands = case Reason of
+		normal -> Commands0;
+		shutdown -> Commands0;
+		{shutdown, _} -> Commands0;
+		_ -> [{log, error,
+				"Ranch listener ~p, connection process ~p, stream ~p "
+				"had its request process ~p exit with reason "
+				"~999999p~n",
+				[Ref, self(), StreamID, Pid, Reason]}
+			|Commands0]
+	end,
+	do_info(StreamID, Exit, [
+		{error_response, 500, #{<<"content-length">> => <<"0">>}, <<>>}
+	|Commands], State);
 %% Request body, auto mode, no body buffered.
 info(StreamID, Info={read_body, Pid, Ref, auto, infinity}, State=#state{read_body_buffer= <<>>}) ->
 	do_info(StreamID, Info, [], State#state{
--- a/test/handlers/crash_h.erl
+++ b/test/handlers/crash_h.erl
@@ -13,4 +13,10 @@ init(_, no_reply) ->
 init(Req, reply) ->
 	_ = cowboy_req:reply(200, Req),
 	ct_helper:ignore(?MODULE, init, 2),
-	error(crash).
+	error(crash);
+init(_, internal_exit) ->
+	ct_helper:ignore(?MODULE, init, 2),
+	exit(internal_exit);
+init(_, external_exit) ->
+	ct_helper:ignore(?MODULE, init, 2),
+	exit(self(), external_exit).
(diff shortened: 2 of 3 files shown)


src/cowboy_stream_h.erl
```diff
@@ -154,6 +154,22 @@ info(StreamID, Exit={'EXIT', Pid, {Reason, Stacktrace}}, State=#state{ref=Ref, p
 	do_info(StreamID, Exit, [
 		{error_response, 500, #{<<"content-length">> => <<"0">>}, <<>>}
 	|Commands], State);
+info(StreamID, Exit={'EXIT', Pid, Reason}, State=#state{ref=Ref, pid=Pid}) ->
+	Commands0 = [{internal_error, Exit, 'Stream process crashed.'}],
```
essen, line 158: Let's use `exited` rather than `crashed`.

src/cowboy_stream_h.erl
```diff
@@ -154,6 +154,22 @@ info(StreamID, Exit={'EXIT', Pid, {Reason, Stacktrace}}, State=#state{ref=Ref, p
...
+		_ -> [{log, error,
+				"Ranch listener ~p, connection process ~p, stream ~p "
+				"had its request process ~p exit with reason "
+				"~999999p~n",
+				[Ref, self(), StreamID, Pid, Reason]}
+			|Commands0]
+	end,
+	do_info(StreamID, Exit, [
+		{error_response, 500, #{<<"content-length">> => <<"0">>}, <<>>}
+	|Commands], State);
```
essen, line 172: You can probably extract all this into a separate function, with slightly different behavior depending on the presence of a stacktrace (use `undefined` for the exit case).

test/handlers/crash_h.erl
```diff
@@ -13,4 +13,10 @@ init(_, no_reply) ->
...
 	_ = cowboy_req:reply(200, Req),
 	ct_helper:ignore(?MODULE, init, 2),
-	error(crash).
+	error(crash);
+init(_, internal_exit) ->
+	ct_helper:ignore(?MODULE, init, 2),
+	exit(internal_exit);
+init(_, external_exit) ->
+	ct_helper:ignore(?MODULE, init, 2),
+	exit(self(), external_exit).
```
essen, line 22: The ignore function is about crashes not exits. This handler is about crashes as well. I don't think this belongs here. A better place for this test would be in stream_handler_SUITE similar to this test https://github.com/ninenines/cowboy/blob/master/test/stream_handler_SUITE.erl#L341-L362 (just get the pid and exit the process immediately, the test shouldn't be that long). It's OK to only test `exit/2` in this case.

test/metrics_SUITE.erl
```diff
@@ -542,3 +544,118 @@ error_response_after_reply(Config) ->
...
+				exit := ProcExit,
+				reason := external_exit
+			}}] = maps:to_list(Procs),
+			true = ProcSpawn =< ProcExit,
+			%% Confirm other metadata are as expected.
+			#{
+				ref := _,
+				pid := From,
+				streamid := 1,
+				reason := {internal_error, {'EXIT', _Pid, external_exit}, 'Stream process crashed.'},
```
essen, line 653: That said, checking this still seems valuable. So I would leave this test. I would just do the exit from the test case instead.

essen (changes requested):
There currently is no mechanism to ignore exit messages. Feel free to send a PR for ct_helper or to just leave the exit message for now. If a PR then please don't spend too much time before proposing a solution, I haven't done any thinking about this yet.

test/metrics_SUITE.erl
hubertlepicki: alright makes sense. On it.

essen:
I have done a simpler implementation that is now in master to be released as Cowboy 2.13. Closing, thanks!

PR #1522, opened 2021-05-29
Fix set-cookie response value check

okeuday:
* Allow a single set-cookie binary as a response header value
  (assuming the value is a list occurs when it isn't a binary)

### DIFF
--- a/src/cowboy_http.erl
+++ b/src/cowboy_http.erl
@@ -1216,7 +1216,7 @@ commands(State, StreamID, [{push, _, _, _, _, _, _, _}|Tail]) ->
 	commands(State, StreamID, Tail).
 
 %% The set-cookie header is special; we can only send one cookie per header.
-headers_to_list(Headers0=#{<<"set-cookie">> := SetCookies}) ->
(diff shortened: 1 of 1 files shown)


essen:
I don't think this is a good idea because single header values can legitimately be iolists.

okeuday:
@essen Yes, the source code was written to only accept iolists for the `set-cookie` header value while other header values are able to be iodata.  The fix was to ensure all header values were able to be iodata, to be more consistent.

essen:
No the way it works is that `set-cookie` as a header breaks the HTTP spec, and there has to be one `set-cookie` header line per cookie. It can't be concatenated like other HTTP headers. So Cowboy has to handle `set-cookie` specially and internally Cowboy treats the `set-cookie` value as a list of cookies. Individual cookies are `iodata()`. You should not treat `set-cookie` as a normal header but instead call `set_resp_cookie`.

There's a separate ticket that I will work on soon that will change the code to explicitly reject a `set-cookie` header and redirect the user to the appropriate function, and I think another to make it simpler for proxies to convert their responses into appropriate calls.

PR #1536, opened 2021-08-29
Fix typos

kianmeng:


kianmeng:
@essen Any updates on this?

essen:
Merged locally, will push later. I have removed the changes that were not documentation with the exception of the one test description since it's logged. I do not want to pollute the history of individual files just for a few comments. Also modifications to erlang.mk, erlmarkdown and whatnot are not relevant to us since these are just copied from elsewhere.

Closing, thanks!

PR #1545, opened 2021-11-04
Add keepalive function that resets the inactivity timer

Jolmberg:
Can be used in a streaming loop when we expect more data to arrive
and want to prevent an inactivity timeout.

(And yes, keepalive has another very specific meaning in the http world so this is a bit of an annoying name clash, but it's also typically what a message like this would be called so I couldn't think of anything better, feel free to change it or suggest something else.)


essen:
This should have a test to demonstrate the problem. I also think this is what we have `set_options` for. It might be better to just add support for the inactivity timeout to `set_options`. Alternatively if all you need is a reset, sending an empty `set_options` will work and doesn't require any change in Cowboy.

Closing, thanks!

PR #1550, opened 2021-11-16
Remove official IRC channel

Maria-12648430:
#ninenines is gone, you said 😓 And freenode is down the gutter... 🙄


essen:
Merged, will push soon. Thanks!

PR #1559, opened 2022-01-07
Get peer cert inderectly through ranch_transport:peercert/1

juise:
The related PR in ranch https://github.com/ninenines/ranch/pull/337


essen:
Seems like I need to merge in Ranch first, wait a few years(?) or 3.0, and then we can do this in Cowboy.

PR #1572, opened 2022-06-06
Minor bugfix in sample code multipart.asciidoc

ahron1:
For reading a multipart message where the part is a file :-
 
this line below leads to a badmap error after the last chunk has been read. So the functionality isn't affected but it just throws an error at the very end.
            `Req = case cow_multipart:form_data(Headers) of`
           
Removing the pattern matching and replacing it with just
            `case cow_multipart:form_data(Headers) of`
           
resolves the issue. Doing this means the recursive call `multipart(Req)` should be replaced with `multipart(Req1)`.

### DIFF
--- a/doc/src/guide/multipart.asciidoc
+++ b/doc/src/guide/multipart.asciidoc
@@ -98,14 +98,14 @@ part is a file:
 multipart(Req0) ->
     case cowboy_req:read_part(Req0) of
         {ok, Headers, Req1} ->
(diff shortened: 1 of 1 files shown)


essen:
I will review the code for a possible fix but this fix isn't correct as you are discarding the Req from inside the case block.

essen:
I am not finding an error. A way to reproduce would help, thank you! Note that there is a simpler version of this code in tests and that hasn't showed any issues. Closing for now.

PR #1575, opened 2022-07-25
Translation into Simplified Chinese for Readme

HariiHe:
Translate Readme.asciidoc into Simplified Chinese for Chinese users


essen:
Feel free to translate documentation in a separate repository or website. I do not have the bandwidth to update multiple documents at this time, and it's not right to replace the English with Chinese. Thanks!

PR #1577, opened 2022-08-02
Create SECURITY.md

JamieSlome:
To let the repository confirm contact@ninenines.eu as its security contact.


essen:
Should have said in the other ticket, but no thanks. I do want a ticket to be opened first regardless of the type of issue.

Issue #1582, opened 2022-10-10
Possibly incorrect behavior on pipelined request shutdown


dmitrivereshchagin:
Current behavior of `cowboy_http:initiate_closing/2` seems strange to me. It calls `cowboy_stream:terminate/3` for all streams in `#state.streams` except the first one. It seems like the intention was to terminate all but the stream currently being processed. This is not guaranteed as the first stream in this list is the last created one.

`cowboy_stream:terminate/3` is also called from `cowboy_http:terminate/2`. Thus it can be called twice for some streams. It makes `cowboy_stream_h` to call metrics callback twice for the same request.

Another possible issue is that `cowboy_http:stream_next/1` doesn't check if the `last_streamid` has been reached. This may cause next response to be written after Cowboy sends "close" connection option. Seems like Gun hides this in tests. It's possible to reproduce it using Netcat.

essen:
This was merged. Closing, thanks!

PR #1583, opened 2022-10-10
Fix shutdown for HTTP/1.1 pipeline

dmitrivereshchagin:
Draft fix for #1582.

* Call `cowboy_stream:terminate/3` once for each stream (do not terminate streams from `cowboy_http:initiate_closing/2`).
* Prevent next response to be written after "close" option has been sent (check if the last stream ID is reached in `cowboy_http:stream_next/1`).

### DIFF
--- a/src/cowboy_http.erl
+++ b/src/cowboy_http.erl
@@ -1445,10 +1445,12 @@ early_error(StatusCode0, #state{socket=Socket, transport=Transport,
 
 initiate_closing(State=#state{streams=[]}, Reason) ->
 	terminate(State, Reason);
-initiate_closing(State=#state{streams=[_Stream|Streams],
+initiate_closing(State=#state{streams=Streams,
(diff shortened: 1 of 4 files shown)


essen:
Isn't a better fix to terminate immediately all streams except the one currently executing (`lists:keytake` would give us that stream and the remainder to terminate) and then store only that remaining stream in `streams` in the state.

In any case a test case reproducing the issue should come first. `http_SUITE:graceful_shutdown_connection` already does some pipelining but the test can likely be enhanced to ensure we get the correct response.

dmitrivereshchagin:
Thanks for your suggestion. Please take a look again.

essen:

dmitrivereshchagin:
Rebased onto master.

essen:
This was merged with minor tweaks. Closing, thanks!

PR #1585, opened 2022-11-01
Feature: add functions to add + remove dispatch routes

artman41:
Adds helper functions to insert & remove dispatch routes at runtime.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 PROJECT = cowboy
 PROJECT_DESCRIPTION = Small, fast, modern HTTP server.
-PROJECT_VERSION = 2.9.0
+PROJECT_VERSION = 2.9.1
 PROJECT_REGISTERED = cowboy_clock
 
 # Options.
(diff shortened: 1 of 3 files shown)


essen:
This changes the route order though?

artman41:
@essen my thought behind it is that if you already had a match-all dispatch `/[...]` then any additional dispatches would fail to match as you'd fall into the first case - with that logic, it makes sense to add newer dispatches before the existing ones.

Happy to modify the implementation to put new dispatches at the end if you think that'd be better though

essen:
I question the approach. I think if something was to be added it would have to be something more general, like a function that traverses routes and allows inserting/removing routes at any point. Go through the list in order and allow returning `[]`, one route tuple, or a list of routes tuples, and use that return value to update the list. Probably worth also adding a way to stop while keeping all subsequent routes as well.

artman41:
@essen that sounds a lot more involved than the easy MR I had in mind 😉 

If you get chance to write up a bit of a detailed idea of what you mean then I'll give it a go 👍

PR #1590, opened 2022-11-22
Respect a return value of Transport:send/2

### LINKED ISSUE #1589: Connections hang in CLOSED state
We have faced an issue when connections (ports and processes owning them) hang forever (we use these connections for infinite streams of events).
```
...
4071 inet_tcp 249       69938937   <0.15260.394>                   *:*                    CLOSED       STREAM
4072 inet_tcp 249       141447159  <0.17014.480>                   *:*                    CLOSED       STREAM
4078 inet_tcp 249       176131209  <0.6376.600>                    *:*                    CLOSED       STREAM
4084 inet_tcp 249       80625928   <0.22290.881>                   *:*                    CLOSED       STREAM
4087 inet_tcp 249       101426932  <0.2095.695>                    *:*                    CLOSED       STREAM
4095 inet_tcp 249       147349243  <0.23210.849>                   *:*                    CLOSED       STREAM
```
Our investigation showed it happens due to combination of configuration, error handling from cowboy side and inet_drv behavior. The interesting thing here is when `send_timeout_close` configuration parameter is used and `inet_drv` closes a connection due to timeout it does not send `tcp_closed` or `tcp_error` message to a port owner. The only thing we can do in this case is always check a return value of `Transport:send/2`.
UPD: I've made a test for this case in #1590

Yozhig:
fixes #1589

### DIFF
--- a/src/cowboy_http.erl
+++ b/src/cowboy_http.erl
@@ -254,7 +254,12 @@ loop(State=#state{parent=Parent, socket=Socket, transport=Transport, opts=Opts,
 			sys:handle_system_msg(Request, From, Parent, ?MODULE, [], State);
 		%% Messages pertaining to a stream.
 		{{Pid, StreamID}, Msg} when Pid =:= self() ->
-			loop(info(State, StreamID, Msg));
+			try info(State, StreamID, Msg) of
+				NewState ->
+					loop(NewState)
+			catch throw:{send, Error} ->
+				terminate(State, Error)
+			end;
 		%% Exit signal from children.
 		Msg = {'EXIT', Pid, _} ->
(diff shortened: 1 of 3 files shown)


essen:

essen:
Hello! I have done a more complete patch covering all the Transport related errors as well as adding a test for HTTP/2 and fixing HTTP/2 as well. I have kept the test commit and added the HTTP/2 test there, then added another commit with the fixes. This also involved adding some information to the documentation because this impacts how stream handler commands are handled (they can now be interrupted by socket errors). Closing, thanks!

Yozhig:
I can't test it right now but it looks good to me. Thank you!

PR #1599, opened 2023-01-31
HTTP/3

essen:
Obviously a work in progress. It's mostly done now. Experimenting is welcome.

Building: it's complicated. Either figure it out or wait for something more stable.


essen:
The most recent push gets us very close from merging to master. I will be unavailable for a few days though so we'll see when that happens.

essen:
It's now as green as it can currently get. I will therefore merge the HTTP/3 work tomorrow after squashing everything in one and performing final checks.

essen:
Merged!!

Issue #1596, opened 2023-01-28
Dependency mismatch between cowboy and gun 2.0


polvalente:
Currently, `gun` 2.0 depends on `cowlib` 2.12.0, but `cowboy` 2.9 depends on `cowlib` 2.11.
This makes it impossible for `gun` to be updated to 2.0 on an Elixir project that also depends on `cowboy` (`elixir_grpc` in this case).

Setting `override: true` is not an option because as a library, we cannot publish to Hex.pm with `:override` specified in the mix.exs file.

Could we get either an updated `cowboy` or a rollback on the requirement in `gun`?

polvalente:
If there was a way to publish cowboy 2.9.1 with the requirement relaxed to the equivalent of Mix's `~> 2.11` that would also be acceptable, but I do not know it that's a possibility in Rebar

essen:
I'm not using Rebar anyway. Right now I can't do what you're asking, the requirement in Gun is because it really requires this version of Cowlib. The `~>` thing is not available either and I don't think that would include 2.12 anyway which Gun requires.

So your only option is to wait for a Cowboy update and not update Gun in the meantime I suppose, or see with the Hex people about removing some of their checks and restrictions which are a bit much IMO.

Out of curiosity which library is this?

polvalente:
It's https://github.com/elixir-grpc/grpc

essen:
Closing in favor of https://github.com/ninenines/cowboy/issues/1591 where I am working on fixing this issue. There's an open question on there you may want to answer. Thank you!

PR #1601, opened 2023-02-10
Update cowlib to 2.12.0

carrascoacd:
## Context

Gun 2.0 has been released. As a consequence, the dependency resolution when using https://github.com/elixir-grpc/grpc is broken.

In order to remove the temporal fork of `gun` and use the official `gun` library, we want to make sure that:
1. `gun` depends on `cowlib` 2.12.0
2. `cowboy` depends on `cowlib` 2.12.0
3. `grpc` depends indirectly on on `cowlib` 2.12.0

@essen @polvalente, this is related to https://github.com/elixir-grpc/grpc/pull/301/files

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,4 +1,4 @@
 {deps, [
-{cowlib,".*",{git,"https://github.com/ninenines/cowlib","2.11.0"}},{ranch,".*",{git,"https://github.com/ninenines/ranch","1.8.0"}}
+{cowlib,".*",{git,"https://github.com/ninenines/cowlib","2.12.0"}},{ranch,".*",{git,"https://github.com/ninenines/ranch","1.8.0"}}
 ]}.
 {erl_opts, [debug_info,warn_export_vars,warn_shadow_vars,warn_obsolete_guard,warn_missing_spec,warn_untyped_record]}.


polvalente:
This is also related to https://github.com/ninenines/cowboy/issues/1596

essen:
Yes but that won't work in the long run, I can't really make a new Cowboy release every time there's a Gun release and vice versa.

Just to be clear, the error ultimately comes from Mix which doesn't allow mismatching dependency versions, right? So we would need the rebar.config file (and Hex metadata) to have the loosy version instead of the strict version. I was thinking of adding a way to override the version in the Erlang.mk configuration and both Rebar and Mix would then get the loose version. But that's only assuming Mix uses the rebar.config for Cowboy.

carrascoacd:
@essen There is another approach that could be simpler, why don't we relax the dependencies? ie using `~> 2.12`. It will allow you to use from `2.12.0` to `2.99.x`. If you use Semver, you shouldn't need to care about minor and bugfix versions. In fact, you want to have them as soon as possible.

essen:
See my previous comment.

carrascoacd:
I don't get you, sorry. According to the doc you can do it:

> For convenience reasons (and because hex.pm mandates semver) hex dependencies can be specified using semver-like syntax:

```erlang
{deps,[
  rebar,                   % fetches latest known version, ignoring pre-releases
  {rebar, "~> 2.0.0"},     % >= 2.0.0 and < 2.1.0`
  {rebar, "~> 2.1.2"},     % >= 2.1.2 and < 2.2.0`
  {rebar, "~> 2.1.3-dev"}` % >= 2.1.3-dev and < 2.2.0`
  {rebar, "~> 2.0"}`       % >= 2.0.0 and < 3.0.0`
  {rebar, "~> 2.1"}`       % >= 2.1.0 and < 3.0.0`
]}.
```

https://rebar3.org/docs/configuration/dependencies/

carrascoacd:
In any case @essen 😁. Since adding lazy dependency is not a requirement, we can merge this and think about how to improve it later if you prefer. Does it make sense?

essen:
I'm not using Rebar, but Erlang.mk. So when Erlang.mk generates the rebar.config for compatibility I am saying it should have an option to override the version. Before I make a new Cowboy version I want to have that done.

I do not want the Erlang.mk version to be fuzzy as only the strict versions were tested and that's all I can guarantee works.

jbernardo95:
cowlib has been update on master btw: https://github.com/ninenines/cowboy/commit/4958af574529464aac797f830e77e742eca74964

essen:
I've had to release Cowboy 2.10.0 for OTP-26 compatibility. I will also make a new Gun version for the same reason, so your issue will be gone for now, but a more long term solution will likely not be done before Cowboy 3.0 (so let's keep this ticket open). I'll ping when Gun is updated.

essen:
OK there's a ticket already so I'm closing this. Expect a new Gun release soon.

Issue #1595, opened 2023-01-25
The file_server example cannot handle unicode filename


linchengcs:
`echi "hello" > priv/中文.html`
`make run`
`curl http://localhost:8080/中文.html`

Then the server crashes

essen:
Please send a patch if you can, shouldn't be too difficult to fix.

essen:
Done in 4f26d6a5730685d00ed1533fe33dc2eb3a51a4e8. Closing, thanks!

PR #1602, opened 2023-02-10
Add UTF-8 support to the file_server example

djankovic:
Fixes https://github.com/ninenines/cowboy/issues/1595


essen:
Thanks I will test and include it later on as we get close to the next version.

essen:
Merged and tweaked using both PRs and some additional bits. Result is in 4f26d6a5730685d00ed1533fe33dc2eb3a51a4e8. Closing, thanks!

PR #1605, opened 2023-02-17
Reject request with invalid Connection header format

0x42:
Cowboy reject request with invalid Connection header format instead of [throwing exception](https://github.com/ninenines/cowlib/issues/129).


zinid:
@essen do we need to improve anything in the PR? We need this fix badly.

Regarding failed CI checks: I think this is unrelated to the changes. Correct me if I wrong :)

essen:

0x42:
Unfortunately CI failed (((
 like here https://github.com/ninenines/cowboy/actions/runs/7114039692/job/19367589145

essen:
The failure on OTP master is expected as there's a bug in OTP causing it. The one on 26.1 is not but likely a flake so I don't think there's any problem. Cheers.

essen:
Merged with tweaks and an additional test for HTTP/1.0. Closing, thanks!

PR #1609, opened 2023-03-31
Add cowboy_decompress_h stream handler

### LINKED ISSUE #946: Transparently handle gzipped request bodies
I have a server that regularly receives relatively large requests, and it would be very helpful if Cowboy accepted compressed requests.

Perhaps look for a `Content-Encoding: gzip` request header, and call `zlib:gunzip` on the response body after it's been fully read from the client. Is this possible?

jdamanalo:
Closes #946.

I added documentation with tentative 2.X as the version where this module is introduced. Alternatively, the documentation could also be left out much like when the other stream handlers were still experimental.

### DIFF
--- a/src/cowboy_decompress_h.erl
+++ b/src/cowboy_decompress_h.erl
@@ -1,0 +1,217 @@
+-module(cowboy_decompress_h).
+-behavior(cowboy_stream).
+
+-export([init/3]).
+-export([data/4]).
+-export([info/3]).
+-export([terminate/3]).
+-export([early_error/5]).
+
+-record(state, {
+	next :: any(),
+	ratio_limit :: non_neg_integer() | undefined,
+	ignore = false :: boolean(),
+	compress = undefined :: undefined | gzip,
+	inflate = undefined :: undefined | zlib:zstream(),
+	is_reading = false :: boolean(),
+	read_body_buffer = <<>> :: binary(),
+	read_body_is_fin = nofin :: nofin | {fin, non_neg_integer()}
+}).
+
+-spec init(cowboy_stream:streamid(), cowboy_req:req(), cowboy:opts())
+	-> {cowboy_stream:commands(), #state{}}.
+init(StreamID, Req, Opts) ->
+	RatioLimit = maps:get(decompress_ratio_limit, Opts, 20),
+	Ignore = maps:get(decompress_ignore, Opts, false),
+	State = check_req(Req),
+	Inflate = case State#state.compress of
+		undefined ->
+			undefined;
+		gzip ->
+			Z = zlib:open(),
+			zlib:inflateInit(Z, 31),
+			Z
+	end,
+	{Commands, Next} = cowboy_stream:init(StreamID, Req, Opts),
+	fold(Commands, State#state{next=Next, ratio_limit=RatioLimit, ignore=Ignore,
+		inflate=Inflate}).
+
+-spec data(cowboy_stream:streamid(), cowboy_stream:fin(), cowboy_req:resp_body(), State)
+	-> {cowboy_stream:commands(), State} when State::#state{}.
+data(StreamID, IsFin, Data, State=#state{next=Next0, inflate=undefined}) ->
+	{Commands, Next} = cowboy_stream:data(StreamID, IsFin, Data, Next0),
+	fold(Commands, State#state{next=Next, read_body_is_fin=IsFin});
+data(StreamID, IsFin, Data, State=#state{next=Next0, ignore=true, read_body_buffer=Buffer}) ->
+	{Commands, Next} = cowboy_stream:data(StreamID, IsFin,
+		<< Buffer/binary, Data/binary >>, Next0),
+	fold(Commands, State#state{next=Next, read_body_is_fin=IsFin});
+data(StreamID, IsFin, Data, State0=#state{next=Next0, ratio_limit=RatioLimit,
+		inflate=Z, is_reading=true, read_body_buffer=Buffer0}) ->
+	Buffer = << Buffer0/binary, Data/binary >>,
+	case inflate(Z, RatioLimit, Buffer) of
+		{error, Type} ->
+			Status = case Type of
+				data -> 400;
+				size -> 413
+			end,
+			Commands = [
+				{error_response, Status, #{<<"content-length">> => <<"0">>}, <<>>},
+				stop
+			],
+			fold(Commands, State0#state{inflate=undefined});
+		{ok, Inflated} ->
+			State = case IsFin of
+				nofin ->
+					State0;
+				fin ->
+					zlib:inflateEnd(Z),
+					zlib:close(Z),
+					State0#state{inflate=undefined}
+			end,
+			{Commands, Next} = cowboy_stream:data(StreamID, IsFin, Inflated, Next0),
+			fold(Commands, State#state{next=Next, read_body_buffer= <<>>,
+				read_body_is_fin=IsFin})
+	end;
+data(_, IsFin, Data, State=#state{read_body_buffer=Buffer0}) ->
+	Buffer = << Buffer0/binary, Data/binary >>,
+	{[], State#state{read_body_buffer=Buffer, read_body_is_fin=IsFin}}.
+
+-spec info(cowboy_stream:streamid(), any(), State)
+	-> {cowboy_stream:commands(), State} when State::#state{}.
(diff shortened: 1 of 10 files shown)


essen:

essen:
Just to be sure since copyright is important: no problem with me adding this to the top of the file?

``` patch
diff --git a/src/cowboy_decompress_h.erl b/src/cowboy_decompress_h.erl
index ffbec25..f256223 100644
--- a/src/cowboy_decompress_h.erl
+++ b/src/cowboy_decompress_h.erl
@@ -1,3 +1,17 @@
+%% Copyright (c) 2023, jdamanalo <jamanalo5@up.edu.ph>
+%%
+%% Permission to use, copy, modify, and/or distribute this software for any
+%% purpose with or without fee is hereby granted, provided that the above
+%% copyright notice and this permission notice appear in all copies.
+%%
+%% THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+%% WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+%% MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+%% ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+%% WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+%% ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+%% OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+
 -module(cowboy_decompress_h).
 -behavior(cowboy_stream).

**Comment by jdamanalo** [comment]
Sure, no problem! joshuadavid.agustin@manalo.ph for my email should be more long-lived.

File: src/cowboy_decompress_h.erl [hunk 1]
```diff
@@ -0,0 +1,217 @@
...
+	catch
+		error:data_error ->
+			zlib:close(Z),
+			{error, data}
+	end.
+
+do_inflate(Z, Size, Limit, Status, _) when Size > Limit ->
+	case Status of
+		continue -> ok;
+		finished -> zlib:inflateEnd(Z)
```
essen, line 208: Do you think this is useful, can't we just zlib:close?

essen: Yeah we can just close. https://github.com/erlang/otp/blob/master/erts/emulator/nifs/common/zlib_nif.c#L288-L293

src/cowboy_decompress_h.erl
```diff
@@ -0,0 +1,217 @@
...
+	{Commands, Next} = cowboy_stream:info(StreamID, Info, Next0),
+	fold(Commands, State#state{next=Next});
+info(StreamID, Info={CommandTag, _, _, _, _}, State=#state{next=Next0, read_body_is_fin=IsFin})
+		when CommandTag =:= read_body; CommandTag =:= read_body_timeout ->
+	{Commands0, Next1} = cowboy_stream:info(StreamID, Info, Next0),
+	{Commands, Next} = data(StreamID, IsFin, <<>>, State#state{next=Next1, is_reading=true}),
+	fold(Commands ++ Commands0, Next);
+info(StreamID, Info={set_options, Opts}, State=#state{next=Next0,
+		ignore=Ignore0, ratio_limit=RatioLimit0}) ->
+	Ignore = maps:get(decompress_ignore, Opts, Ignore0),
```
essen, line 91: Second thoughts on this. We probably shouldn't allow setting `ignore` in the middle of the data, right? If we do the data becomes useless but is still passed forward.

src/cowboy_decompress_h.erl
```diff
@@ -0,0 +1,217 @@
...
+							Codings = build_accept_encoding(WithGzip),
+							Headers#{<<"accept-encoding">> => Codings};
+						{_, _} ->
+							Headers;
+						false ->
+							Headers#{<<"accept-encoding">> => [AcceptEncoding, <<", gzip">>]}
+					end
+			end
+	catch _:_ ->
+		Headers#{<<"accept-encoding">> => <<"gzip">>}
```
essen, line 170: I don't know if that's the right approach, overriding the header completely. Maybe this is a case where we would be better off crashing: if you send a response with a broken accept-encoding header and use this stream handler, you deserve the crash IMO.

src/cowboy_decompress_h.erl
```diff
@@ -0,0 +1,217 @@
...
+	Headers#{<<"accept-encoding">> => <<"gzip">>}.
+
+%% From cowlib, maybe expose?
+qvalue_to_iodata(0) -> <<"0">>;
+qvalue_to_iodata(Q) when Q < 10 -> [<<"0.00">>, integer_to_binary(Q)];
+qvalue_to_iodata(Q) when Q < 100 -> [<<"0.0">>, integer_to_binary(Q)];
+qvalue_to_iodata(Q) when Q < 1000 -> [<<"0.">>, integer_to_binary(Q)];
+qvalue_to_iodata(1000) -> <<"1">>.
+
+build_accept_encoding([{ContentCoding, Q}|Tail]) ->
```
essen, line 182: I'll add to Cowlib instead.

essen (commented):
I won't have time to finish today so I'll leave a few questions. I'll continue early January.

Overall this is solid. Great job! Thanks. I did a few simplifications and whatnot, with the biggest change being making `read_body_buffer` a list of binaries, to avoid doing unneeded binary concatenations and memory allocations. We can give this list of binaries directly to the inflate functions and they won't need to convert them to a single binary. Win win. We do convert when passing the data forward to the next stream handler (whether we inflate or not).

I haven't reviewed the tests yet but after a quick look you get points for completeness.

Merry Christmas, cheers.

src/cowboy_decompress_h.erl
essen: OK after some testing we really shouldn't allow setting the `ignore` option after reading has begun. We can still allow `set_options` to set it, but only if there hasn't been data going through yet. I'll fix it.

src/cowboy_decompress_h.erl
essen: I'll leave a todo and add to Cowlib in a future release. I'll avoid doing Cowlib changes this time around.

src/cowboy_decompress_h.erl
essen: Should be done locally.

src/cowboy_decompress_h.erl
essen: I'm thinking that we shouldn't crash but we should not overwrite the accept-encoding header the user has set. If the user is doing something weird then they can include gzip themselves. This is very much an edge case anyway so let's go with that for now and see later if that causes problems.

essen: Hm but we already do replace in cowboy_compress_h when a header is invalid so I guess we can keep this same behavior here. No change needed here in that case.

essen:
I've done the above changes and renamed the option to `decompress_enabled`. I am updating the documentation now. All that's left after that is checking the tests, perhaps consolidating a thing or two, and we're good.

test/decompress_SUITE.erl
```diff
@@ -0,0 +1,327 @@
...
+	Body = zlib:gzip(Data),
+	{200, Headers, Data} = do_post("/accept-implicit-header",
+		[{<<"content-encoding">>, <<"gzip">>}], Body, Config),
+	{_, <<"identity, *;q=0.5">>} = lists:keyfind(<<"accept-encoding">>, 1, Headers),
+	ok.
+
+%% RFC 9110. Section 12.5.3. 10.1. If no Accept-Encoding header field is in the
+%% request, any content coding is considered acceptable by the user agent.
+%%
+%% Don't add anything on error or when that handler is not used.
```
essen, line 311: Just some info: this comment refers to how user agents are behaving, there is nothing to do here as far as Cowboy is concerned. It could only be relevant to the compression stream handler, not the decompression one.

essen:
OK I am pushing the changes. A review would be appreciated, my changes are in fd9711d9495e4ddcd41eda7a284dfc7f37c11f15. Closing, thanks!

jdamanalo:
Reviewed and everything seems in order to me. I learned a lot from your extensive guidance. Thanks again, I really appreciate it! 🙌

PR #1614, opened 2023-09-10
Fix typo

cmnstmntmn:


essen:
This is the wrong place, Erlang.mk packages will be removed at some point in the future, and this was copied automatically from the repository.

PR #1617, opened 2023-10-31
Add 'max_cancel_stream_rate' config for the rapid reset attack

### LINKED ISSUE #1615: Is Cowboy affected by the HTTP/2 Rapid Reset attack?
See: https://nvd.nist.gov/vuln/detail/CVE-2023-44487

Overview: https://blog.cloudflare.com/zero-day-rapid-reset-http2-record-breaking-ddos-attack/

zuiderkwast:
Fixes #1615

### DIFF
--- a/doc/src/manual/cowboy_http2.asciidoc
+++ b/doc/src/manual/cowboy_http2.asciidoc
@@ -201,8 +201,8 @@ similar to a supervisor restart intensity/period.
 
 max_cancel_stream_rate ({500, 10000})::
 
-Maximum cancel stream rate per connection. This can be used to protect
-against the rapid reset attack (CVE-2023-44487), by limiting the
+Maximum cancel stream rate per connection. This can be used to
+protect against misbehaving or malicious peers, by limiting the
 number of streams that the peer can reset over a certain time period.
 The rate is expressed as a tuple `{NumCancels, TimeMs}`. This is
 similar to a supervisor restart intensity/period.
--- a/doc/src/manual/cowboy_http2.asciidoc
+++ b/doc/src/manual/cowboy_http2.asciidoc
@@ -39,6 +39,7 @@ opts() :: #{
     max_frame_size_sent            => 16384..16777215 | infinity,
     max_received_frame_rate        => {pos_integer(), timeout()},
     max_reset_stream_rate          => {pos_integer(), timeout()},
+    max_cancel_stream_rate         => {pos_integer(), timeout()},
     max_stream_buffer_size         => non_neg_integer(),
     max_stream_window_size         => 0..16#7fffffff,
     preface_timeout                => timeout(),
@@ -196,6 +197,14 @@ protect against misbehaving or malicious peers that do not follow
 the protocol, leading to the server resetting streams, by limiting
 the number of streams that can be reset over a certain time period.
 The rate is expressed as a tuple `{NumResets, TimeMs}`. This is
+similar to a supervisor restart intensity/period.
+
+max_cancel_stream_rate ({500, 10000})::
+
+Maximum cancel stream rate per connection. This can be used to protect
+against the rapid reset attack (CVE-2023-44487), by limiting the
+number of streams that the peer can reset over a certain time period.
+The rate is expressed as a tuple `{NumCancels, TimeMs}`. This is
 similar to a supervisor restart intensity/period.
 
 max_stream_buffer_size (8000000)::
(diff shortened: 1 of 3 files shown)


src/cowboy_http2.erl
```diff
@@ -189,6 +196,12 @@ init_reset_rate_limiting(State=#state{opts=Opts}, CurrentTime) ->
 		reset_rate_num=ResetRateNum, reset_rate_time=add_period(CurrentTime, ResetRatePeriod)
 	}.
 
+init_cancel_rate_limiting(State=#state{opts=Opts}, CurrentTime) ->
+	{CancelRateNum, CancelRatePeriod} = maps:get(max_cancel_stream_rate, Opts, {100, 10000}),
```
bjosv, line 200:
```suggestion
	{CancelRateNum, CancelRatePeriod} = maps:get(max_cancel_stream_rate, Opts, {500, 10000}),
```
to match docs.

bjosv (commented):
Nice!

doc/src/manual/cowboy_http2.asciidoc
```diff
@@ -198,6 +199,14 @@ the number of streams that can be reset over a certain time period.
 The rate is expressed as a tuple `{NumResets, TimeMs}`. This is
 similar to a supervisor restart intensity/period.
 
+max_cancel_stream_rate ({500, 10000})::
+
+Maximum cancel stream rate per connection. This can be used to protect
+against the rapid reset attack (CVE-2023-44487), by limiting the
```
essen, line 205: Let's keep `misbehaving or malicious peers` instead of referencing the CVE. I don't want people to mistakenly think they are protected against this or other CVEs just because Cowboy has dropped the connection. If necessary we can later on discuss protection against CVEs in a dedicated chapter of the user guide.

src/cowboy_http2.erl
```diff
@@ -1137,6 +1157,21 @@ reset_rate(State0=#state{reset_rate_num=Num0, reset_rate_time=Time}) ->
 			{ok, State0#state{reset_rate_num=Num}}
 	end.
 
+cancel_rate(State0=#state{cancel_rate_num=Num0, cancel_rate_time=Time}) ->
```
essen, line 1160: This function should be just below the function that calls it (line 595).

src/cowboy_http2.erl
```diff
@@ -563,14 +576,21 @@ early_error(State0=#state{ref=Ref, opts=Opts, peer=Peer},
...
-			terminate_stream_handler(State, StreamID, Reason, StreamState),
+			terminate_stream_handler(State0, StreamID, Reason, StreamState),
 			Children = cowboy_children:shutdown(Children0, StreamID),
-			State#state{streams=Streams, children=Children};
+			State0#state{streams=Streams, children=Children};
 		error ->
-			State
+			State0
+	end,
+	case cancel_rate(State1) of
```
essen, line 588: This should only be applied if a stream was indeed reset. It's possible the stream was already gone (so we didn't waste resources) or it's also possible that Cowboy is seen as misbehaving and the client sends more than one RST_STREAM for a single stream. Since the purpose is to protect against actual stream cancellations (and not frame flood, covered by a separate setting), we should only count those actual cancellations.

essen (changes requested):
Thanks!

There should also be tests similar to `http2_reset_flood` in `security_SUITE` except we would have variants of headers+rst_stream sent:

* a test where we send many HEADERS+RST_STREAM immediately
* a test where we send N (10?) HEADERS then N (10?) RST_STREAM, repeatedly
* a test where we send many HEADERS at once and then all RST_STREAM

src/cowboy_http2.erl
zuiderkwast: Done. I renamed it to `cancel_rate_limit` and simplified it a bit by just returning the new state (not wrapped in an ok-tuple) and calling `terminate` directly instead of returning `error`.

doc/src/manual/cowboy_http2.asciidoc
zuiderkwast: Done.

zuiderkwast:
Fixed comments and added a test case. Please look again when you can.

essen:

zuiderkwast:
Awesome that we have a working CI. Good job!

Merge is fine since you'll do a squash-merge anyway, right?

Some test runs (but not all) failed in tracer_SUITE. Do you have a clue?

essen:
tracer_SUITE on master is a bug in OTP, see https://github.com/erlang/otp/issues/7926

I'd prefer a non-merge rebase but don't worry about it I'll deal with it.

essen:
Merged, thanks!

Issue #1619, opened 2023-11-07
Reject responses that have explicit set-cookie


essen:
This causes confusion otherwise. See https://github.com/ninenines/cowboy/issues/1483#issuecomment-1790344344

We want users to call set_resp_cookie or to manually set the raw resp_cookies value in the Req. A function to do that may be welcome. This is mainly useful for proxies though so not sure how much we should do.

geeksilva97:
I am looking for issues in Erlang to practice more. Would you say I could pick this one?

essen:
Hello @geeksilva97 sorry I had not received an email following your comment, guess it got lost. Yes it would be a good choice. Basically the functions `set_resp_header`, `set_resp_headers`, `inform`, `reply`, `stream_reply` and `stream_trailers` should fail if the headers given in argument contain `set-cookie`.

geeksilva97:
Thank you, @essen . Will give it try, starting tomorrow.

geeksilva97:
Just a confirmation, @essen . Should I throw an error? If so, are you expecting anything special?

essen:
Please use something of the form:

```
exit({response_error, invalid_header,
    'The set-cookie header is special and must be set using cowboy_req:set_resp_cookie/3,4.'})
```

See a40bab8fb3491a587c93bedb53404decacfe49dc for a more extensive example.

geeksilva97:
> Please use something of the form:
> 
> ```
> exit({response_error, invalid_header,
>     'The set-cookie header is special and must be set using cowboy_req:set_resp_cookie/3,4.'})
> ```
> 
> See [a40bab8](https://github.com/ninenines/cowboy/commit/a40bab8fb3491a587c93bedb53404decacfe49dc) for a more extensive example.

Thank you. Will do.

essen:
Done. Closing, thank you!

PR #1624, opened 2024-01-09
Reject responses with explicit set-cookie header

geeksilva97:
This PR addresses the issue #1619 .

## Proposed changes

- reject response with explicit `<<"set-cookie">>` in headers

### DIFF
--- a/src/cowboy_req.erl
+++ b/src/cowboy_req.erl
@@ -718,6 +718,9 @@ set_resp_cookie(Name, Value, Req, Opts) ->
 
 -spec set_resp_header(binary(), iodata(), Req)
 	-> Req when Req::req().
+set_resp_header(<<"set-cookie">>, _, _) ->
+	exit({response_error, invalid_header,
+    'The set-cookie header is special and must be set using cowboy_req:set_resp_cookie/3,4.'});
 set_resp_header(Name, Value, Req=#{resp_headers := RespHeaders}) ->
 	Req#{resp_headers => RespHeaders#{Name => Value}};
 set_resp_header(Name,Value, Req) ->
@@ -725,6 +728,10 @@ set_resp_header(Name,Value, Req) ->
 
 -spec set_resp_headers(cowboy:http_headers(), Req)
 	-> Req when Req::req().
+
+set_resp_headers(#{<<"set-cookie">> := _}, _) ->
+	exit({response_error, invalid_header,
+    'The set-cookie header is special and must be set using cowboy_req:set_resp_cookie/3,4.'});
 set_resp_headers(Headers, Req=#{resp_headers := RespHeaders}) ->
 	Req#{resp_headers => maps:merge(RespHeaders, Headers)};
 set_resp_headers(Headers, Req) ->
@@ -781,6 +788,9 @@ inform(Status, Req) ->
 inform(_, _, #{has_sent_resp := _}) ->
 	exit({response_error, response_already_sent,
 		'The final response has already been sent.'});
+inform(_, _, #{<<"set-cookie">> := _}) ->
+	exit({response_error, invalid_header,
+    'The set-cookie header is special and must be set using cowboy_req:set_resp_cookie/3,4.'});
 inform(Status, Headers, Req) when is_integer(Status); is_binary(Status) ->
 	cast({inform, Status, Headers}, Req).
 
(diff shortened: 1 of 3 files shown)


src/cowboy_req.erl
```diff
@@ -781,6 +788,9 @@ inform(Status, Req) ->
 inform(_, _, #{has_sent_resp := _}) ->
 	exit({response_error, response_already_sent,
 		'The final response has already been sent.'});
+inform(_, _, #{<<"set-cookie">> := _}) ->
```
geeksilva97, line 791:
```suggestion
inform(_, #{<<"set-cookie">> := _}, _) ->
```

geeksilva97:
Could you take a look when you have time, @essen ? Thanks in advance.

src/cowboy_req.erl
```diff
@@ -15,6 +15,9 @@
 
 -module(cowboy_req).
 
+-define(INVALID_COOKIE_HEADER_ERROR, exit({response_error, invalid_header,
+    'The set-cookie header is special and must be set using cowboy_req:set_resp_cookie/3,4.'})).
```
essen, line 19: No macros please.

essen:
Yes tests are fine like this, though since there's a crash you should also call `ct_helper:ignore(M, F, A),` before the crash, with MFA where the crash occurs. It will silence the expected crash.

src/cowboy_req.erl
geeksilva97: for sure. will move to a function

essen: Just inline is fine. There isn't too many of them. The messages are likely to be slightly edited later on anyway depending on context.

geeksilva97:
@essen would you have any clue how we can test this on `stream_trailers`?

I tried adding this to `resp_h.erl` - inside of `do(<<"stream_trailers">>, Req0, Opts)`.

```erl
<<"set_cookie">> ->
			ct_helper:ignore(cowboy_req, stream_trailers, 2),
			Req = cowboy_req:stream_reply(200, #{
				<<"trailer">> => <<"grpc-status">>
			}, Req0),
			%% The size should be larger than StreamSize and ConnSize
			cowboy_req:stream_body(<<0:80000000>>, nofin, Req),
			cowboy_req:stream_trailers(#{
				<<"set-cookie">> => <<"name=cormano">>
			}, Req),
			{ok, Req, Opts};
```

Then I tried something like this

```erl
do_trailers("resp/stream_trailers/set_cookie", Config)
```

It broke the test suite - I didn't even check the return

It breaks somewhere else it is also a badmatch.

essen:
Sometimes it can be difficult to test crashes. Some of the Cowboy tests trigger the crash but the test case has no way of confirming it happened. That's OK for the time being.

I will pick this up after I'm done with my current ticket. A quick look seems fine. Thank you!

geeksilva97:
> Sometimes it can be difficult to test crashes. Some of the Cowboy tests trigger the crash but the test case has no way of confirming it happened. That's OK for the time being.
> 
> I will pick this up after I'm done with my current ticket. A quick look seems fine. Thank you!

Sounds good. Thanks for the help. Let me know if I can help with something. 

I will look for another issue - would you have any suggestion?

essen:
Merged with a number of fixes, thank you!

Note that most of the tests were not testing the right crash as you copied test code that used bad arguments to trigger crashes and didn't update them.

josevalim:
Hi @essen! This pull request unfortunately breaks all Elixir projects that depend on Cowboy via Plug, since Plug manages the cookie headers by itself. Is there any chance this could be skipped by using a flag? Or alternatively, would you accept a pull request that adds `set_resp_cookie(HeaderValue, Req :: cowboy_req:req())`, so we can give the whole header value (which we have already accumulated and returned)? Thank you.

essen:
For now you can modify the Req object directly (basically doing what `set_resp_cookie(HeaderValue, ...)` would do).

Not sure this has to be in Cowboy but feel free to open a PR and let's gather feedback.

essen:
For future releases perhaps Plug can add Cowboy master to its CI? Maybe on a schedule, running every week. This would have catched this. I am working immediately on 2.12 and this will take a while (HTTP/3) so getting feedback before release would help.

Issue #993, opened 2016-06-02
REST: always send Allow header


essen:
There's basically no reasons not to send it, it's small and cheap to build. Currently we send it on 405 and OPTIONS responses, but it's probably best to always send it.

geeksilva97:
Yeah, it is me again trying to find an issue.

essen:
Done!!

PR #1625, opened 2024-01-17
Return allow header

geeksilva97:
This PR addresses #993 

### Proposed changes

- Add allow header to all responses
- Minor refactor


geeksilva97:
Hopefully, I could make it right now @essen

essen:
This needs tests. Proper tests would show you this doesn't work: for example if the request ends in `service_available` the allowed methods aren't currently known and the value is `undefined` so your code will crash.

The appropriate way to implement this is probably to do `set_resp_headers` after the `allowed_methods` callback was handled.

geeksilva97:
Will be working on that this week

PR #1504, opened 2021-02-13
match_qs crashes with badkey error

farhadi:
`cowboy_req:match_qs` for constraints given as {Key, Constraint} crashes with `badkey` error if key is missing.
Expected behaviour is to crash with `required` error.

PR #1628, opened 2024-01-18
fix match_qs with constraints when key is not present

geeksilva97:
This PR continues what was started here #1504


essen:
Merged, thanks!

PR #1633, opened 2024-02-03
return allow header

geeksilva97:


src/cowboy_rest.erl
```diff
@@ -328,7 +328,9 @@ uri_too_long(Req, State) ->
 allowed_methods(Req, State=#state{method=Method}) ->
 	case call(Req, State, allowed_methods) of
 		no_call when Method =:= <<"HEAD">>; Method =:= <<"GET">> ->
-			next(Req, State, fun malformed_request/2);
+			Allow = <<"HEAD, GET, OPTIONS">>,
```
geeksilva97, line 331: Am I right to assume this? If the `allowed_methods` was not provided, the default headers will be HEAD, GET, and OPTIONS.

cc @essen

geeksilva97:
Still in progress, I am following the suggestion made in a already close pr

test/rest_handler_SUITE.erl
```diff
@@ -483,7 +484,8 @@ create_resource_created(Config) ->
 	Ref = gun:post(ConnPid, "/create_resource?created", [
 		{<<"content-type">>, <<"application/text">>}
 	], <<"hello">>, #{}),
-	{response, _, 201, _} = gun:await(ConnPid, Ref),
+	{response, _, 201, Headers} = gun:await(ConnPid, Ref),
+	{_, <<"POST">>} = lists:keyfind(<<"allow">>, 1, Headers),
```
geeksilva97, line 488: Used the already existing handlers to check the allow header

geeksilva97:
Sup @essen . Could you share your thoughts?

src/cowboy_rest.erl
```diff
@@ -324,29 +324,33 @@ known_methods(Req, State=#state{method=Method}) ->
...
 %% allowed_methods/2 should return a list of binary methods.
 allowed_methods(Req, State=#state{method=Method}) ->
+	DefaultAllowedMethods = [<<"HEAD">>, <<"GET">>, <<"OPTIONS">>],
 	case call(Req, State, allowed_methods) of
 		no_call when Method =:= <<"HEAD">>; Method =:= <<"GET">> ->
-			next(Req, State, fun malformed_request/2);
+			Allow = stringify_allowed_methods(DefaultAllowedMethods),
+			Req2 = cowboy_req:set_resp_header(<<"allow">>, Allow, Req),
+			next(Req2, State, fun malformed_request/2);
 		no_call when Method =:= <<"OPTIONS">> ->
```
essen, line 339: We should always set the header in this function. I'm not sure it would be set in all cases for OPTIONS otherwise. Make a `set_allow_header` common function that sets the header from here, and remove the header setting from further down the code.

geeksilva97: So, the responsibility of adding the `allow` header will be centralized in this function, right? Sounds good to me.

essen: Of adding the default allow header yes. Then handlers can always override this by setting their own.

geeksilva97: Done!

geeksilva97:
@essen could you look into this PR one more time when you get a chance? Thanks.

geeksilva97:
> @essen could you look into this PR one more time when you get a chance? Thanks.

👀

essen (commented):
Please don't ping repeatedly I always come back eventually.

src/cowboy_rest.erl
```diff
@@ -324,41 +324,43 @@ known_methods(Req, State=#state{method=Method}) ->
 uri_too_long(Req, State) ->
 	expect(Req, State, uri_too_long, false, fun allowed_methods/2, 414).
 
+stringify_allowed_methods(MethodList) when is_list(MethodList) ->
+	case MethodList of
+		[] -> <<>>;
+		_ ->
+		<< ", ", Allow/binary >> = << << ", ", M/binary >> || M <- MethodList >>,
+		Allow
+	end.
```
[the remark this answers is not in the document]
geeksilva97: Okay. I will leave that then.

src/cowboy_rest.erl
```diff
@@ -413,16 +415,10 @@ valid_entity_length(Req, State) ->
 
 %% If you need to add additional headers to the response at this point,
 %% you should do it directly in the options/2 call using set_resp_headers.
-options(Req, State=#state{allowed_methods=Methods, method= <<"OPTIONS">>}) ->
```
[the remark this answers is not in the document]
geeksilva97: done

essen:
I've implemented it a bit differently. Closing, thanks!!

Issue #1295, opened 2018-06-04
Accept tuple lists in set_resp_headers


essen:
To make it easier to deal with different implementations (like Gun).

https://github.com/ninenines/cowboy/pull/1287#event-1659654509

geeksilva97:
Am I right to assume you meant something like the following?

```erl
Req = cowboy_req:set_resp_headers([
    {<<"content-type">> , <<"text/html">>},
    {<<"content-encoding">>, <<"gzip">>}
], Req0).
```

essen:
Yes.

geeksilva97:
Will we still do this [merge](https://github.com/ninenines/cowboy/blob/master/src/cowboy_req.erl#L735)? What should happen in this scenario?

```erl
Req = cowboy_req:set_resp_headers([
    {<<"set-cookie">> , <<"name=essen">>},
    {<<"set-cookie">>, <<"age=28">>}
], Req0).
```

I would say it should call the set_resp_cookie twice. Does it make sense?

essen:
In the case of set-cookie it should not call set_resp_cookie but rather append the cookie to `resp_cookie` (like set_resp_cookie does, but without having to parse then rebuild the cookie).

Because there can be duplicates of other headers, the headers should first be processed a little: set-cookie can be handled here, and then duplicate headers must be consolidated (`Val1 + ", " + Val2`). Then the result can be made into a map and merged (overriding existing headers if there's already been a duplicate set).

It probably should not be named `set_resp_headers` since the behavior differs, but `set_resp_headers_list` or something.

geeksilva97:
Got it

essen:
The goal of this ticket is not to add support for multiple of the same header. Hopefully that's not needed; if it is then we'd need to have a separate interface for this.

You SHOULD be able to have multiple links in one Link header:

```
   Link: </TheBook/chapter2>;
         rel="previous"; title*=UTF-8'de'letztes%20Kapitel,
         </TheBook/chapter4>;
         rel="next"; title*=UTF-8'de'n%c3%a4chstes%20Kapitel
```

https://datatracker.ietf.org/doc/html/rfc8288#section-3.5

But your Link headers are not including the < > so they might not follow the spec in other ways.

essen:
Note that the draft you linked is outdated. The most recent draft has correct Link headers: https://www.ietf.org/archive/id/draft-ietf-wish-whip-13.html#section-4.4

You should be able to use a single Link header to convey the same information.

[...]

PR #1634, opened 2024-02-06
Add set_resp_headers_list

geeksilva97:
This PR addresses #1295


src/cowboy_req.erl
```diff
@@ -366,6 +367,30 @@ uri2_test() ->
 	<<"http://localhost/path?dummy=2785">> = iolist_to_binary(uri(Req, #{port => 80})),
 	<<"https://localhost/path?dummy=2785">> = iolist_to_binary(uri(Req, #{scheme => "https", port => 443})),
 	ok.
+
+resp_headers_test() ->
```
abarbosa-godaddy, line 371: If not need we can remove this unit test. I was using just to validate the behaviour

geeksilva97, line 371: If not need we can remove this unit test. I was using just to validate the behavior

src/cowboy_req.erl
```diff
@@ -726,6 +751,30 @@ set_resp_header(Name, Value, Req=#{resp_headers := RespHeaders}) ->
...
+
+-spec set_resp_headers_list(list({ binary(), iodata() }), Req)
+	-> Req when Req::req().
+set_resp_headers_list(HeaderTupleList, Req) ->
+	set_resp_headers_list(HeaderTupleList, #{}, Req).
+
+set_resp_headers_list([], Map, Req) ->
+	set_resp_headers(Map, Req);
+
+set_resp_headers_list([{<<"set-cookie">>, _} | Headers], Map, Req) ->
```
geeksilva97, line 764: @essen I wonder if to have some consistency with the `set_resp_header` we should instead throw an error.

geeksilva97:
Integration tests will be added soon. I would like to get your thoughts first, @essen

essen:
Probably simpler to `maps:from_list` and then `maps:merge`.

essen:
Nevermind I see now why it's done like this.

essen:
I've done a different implementation. Closing, thanks!!

PR #1641, opened 2024-04-02
Fix cookie filter

maehjam:
The cookie filter function did split the cookies only into two parts before filtering. This was causing problems, when there were more than one cookie before the first matching cookie.

See the adjusted test case: Filtering and parsing the cookies `"bad name=strawberry; another bad name=strawberry; cake=strawberry"` did return `"[]"` before this fix.


essen:
Thank you! Don't worry about CI, PR is probably fine. I likely just messed up something.

essen:
Merged, thanks!

PR #1646, opened 2024-04-11
Cancel timer async

ansd:
to improve performance on the critical path


essen:
Cancel timer is no longer a critical path in Websocket following optimisations, however it remains critical for HTTP so I'm leaving this open for the time being.

essen:
I cannot measure the difference with/without this for HTTP. The best improvement would be to not reset the timer so often. That said, it doesn't hurt not to have this even if the improvement cannot be measured at this time.

essen:
Merged, thanks!

ansd:
Thank you!

PR #1648, opened 2024-05-16
Correcting transport options in `send_timeout_close` tests

bjosv:
While attempting to lift the dependency `ranch` from 1.8.0 to 2.1.0
these faults was indicated due to the improved validation of transport options in 2.1.0.


essen:
Merged, thanks!

PR #1657, opened 2024-12-08
change behavior to behaviour

mustafapc:
fixed behavior in plugins.mk


essen:
Thank you! This is not incorrect, both are accepted and mean the same thing. There's no need to change one for the other.

PR #1660, opened 2025-01-08
fix: cowboy_static mimetypes type

MarkoMin:
I believe that this is missing from the specs according to the documentation and the specs of `cow_mimetypes:web/1`. 

EDIT: I found few more similar underspecs in the same module so I included them in this PR.

### DIFF
--- a/src/cowboy_static.erl
+++ b/src/cowboy_static.erl
@@ -29,7 +29,7 @@
 -type extra_charset() :: {charset, module(), function()} | {charset, binary()}.
 -type extra_etag() :: {etag, module(), function()} | {etag, false}.
 -type extra_mimetypes() :: {mimetypes, module(), function()}
-	| {mimetypes, binary() | {binary(), binary(), [{binary(), binary()}]}}.
+	| {mimetypes, binary() | {binary(), binary(), [{binary(), binary()}] | '*'}}.
 -type extra() :: [extra_charset() | extra_etag() | extra_mimetypes()].
 -type opts() :: {file | dir, string() | binary()}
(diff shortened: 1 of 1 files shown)


src/cowboy_static.erl
```diff
@@ -332,7 +332,7 @@ forbidden(Req, State) ->
 %% Detect the mimetype of the file.
 
 -spec content_types_provided(Req, State)
-	-> {[{binary(), get_file}], Req, State}
+	-> {[{binary() | {binary(), binary(), [{binary(), binary()}] | '*'}, get_file}], Req, State}
```
essen, line 335: You cannot provide `text/html/*`, you can only request it. So this spec fix is incorrect at least. So is the `extra_mimetypes()` change. The other changes look fine.

MarkoMin: If so, then documentation must be wrong or I'm getting something wrong - docs for `content_types_provided` are the following:
```erlang
content_types_provided(Req, State) -> {Result, Req, State}

Result     :: [{binary() | ParsedMime, ProvideCallback :: atom()}]
ParsedMime :: {Type :: binary(), SubType :: binary(), '*' | Params}
Params     :: [{Key :: binary(), Value :: binary()}]

Default     - [{{ <<"text">>, <<"html">>, '*'}, to_html}]
```

essen: OK my bad I haven't looked at this code in too long. This is before negotiation and indicates we match any parameters. So this appears correct after all.

essen:
Merged, thanks!

PR #1663, opened 2025-01-16
Websocket: Allow setting the max_frame_size option dynamically

essen:
This can be used to limit the maximum frame size before some authentication or other validation is completed.


essen:
Merged!!

PR #1665, opened 2025-01-22
Improve websocket compressed

essen:
https://github.com/ninenines/cowlib/issues/139


essen:
Next is adding tests.

PR #1676, opened 2025-04-15
Make dynamic_buffer start at 128 and change less abruptly

essen:
Based on RabbitMQ performance testing.

To be done in the next feature release.


essen:
Merged.

Issue #1054, opened 2016-11-29
Implement a Websocket backpressure mechanism


essen:
See https://github.com/rabbitmq/rabbitmq-web-stomp/issues/40 for details.

Not sure what form it should take exactly but pretty sure it should be something the handler can control. Cowboy should only worry about the handler telling it to read, or not read, from the socket. So having a command to stop/resume reading would be enough.

essen:
Websocket doesn't have flow control which means that such a feature would prevent pings from being responded to. This needs to be clearly explained in the docs.

essen:
I have added `{active, boolean()}` to my local copy. It will be pushed soon but will remain undocumented until Cowboy 2.6.

PR #1685, opened 2025-08-21
Direct data delivery for HTTP/2 Websocket

essen:
The current implementation moves the Websocket data through stream handlers before the Websocket process receives it and parses it. The new mechanism sends the data directly as an Erlang message without any intermediary. It also manages HTTP/2 flow control in a simpler way (it maintains it at a configured value).

A new switch_protocol option (coming from the options returned from init/2) called `data_delivery` decides which mechanism to use: `stream_handlers` is the old behavior; `direct` the new.

WIP: I need to make sure flow control updates act as intended
     and don't increase too much. The default flow control value
	 will depend on that. The value must be configurable.

WIP: Need to test the direct mechanism in other test suites

WIP: the benchmark test suite is greatly modified because I
     wanted to make sure Gun's poor performance was not negatively
	 impacting results (this bit me earlier). Some tests use it,
	 some don't, we should keep both for good measure.

WIP: Other minor todos.


src/cowboy_websocket.erl
```diff
@@ -374,10 +391,11 @@ active(State) ->
 	maybe_read_body(State),
 	State#state{active=true}.
 
-passive(State=#state{transport=undefined}) ->
+passive(State=#state{transport={data_delivery, _}}) ->
 	%% Unfortunately we cannot currently cancel read_body.
 	%% But that's OK, we will just stop reading the body
 	%% after the next message.
+	%% @todo We can't stop relay data_delivery currently.
```
essen, line 398: Stop of reads comes from:

* https://github.com/ninenines/cowboy/issues/1054
* https://github.com/rabbitmq/rabbitmq-web-stomp/issues/40

But to make it work with this feature we need a way to signal that we don't want to continue reading data from the stream.

PR #1688, opened 2025-09-03
Use raw mode to read from the file info

NelsonVides:
Rescue from https://github.com/ninenines/cowboy/pull/1687, don't know why it was closed

See https://erlangforums.com/t/cowboy-2-13-0-performance-bottleneck-at-8-7k-rps-on-erlang-28-0-2/5004.

I've tried to add the original author to the commit as well. Just wanted to put this change into master soon enough :)


essen:
Thanks! I will include it in the next release that I'm already preparing.

essen:
Merged, thanks!

Issue #1686, opened 2025-08-28
Type error in cowboy_rest


ThomasArts:
Cowboy has type specifications for the callbacks in `cowboy_rest` and others, which really helps when using eqwalizer and soing some type checking on applications using cowboy.

The type for `last_modifed` looks like this:
```
 last_modified(Req, State) -> {Result, Req, State}

Result  :: calendar:datetime()
Default  - no last modified value
```

But sometimes one cannot provide a `datetime` and just wants to return `undefined` to disable adding the `LastModified` header.
All the code is there and returning undefined works perfectly well, apart from the fact that it provides a type error for the user code.

The fix seems simple, but there might be deeper reasons why this type is not exported?

```
---
 src/cowboy_rest.erl | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/cowboy_rest.erl b/src/cowboy_rest.erl
index 9f30fcf..13ec89c 100644
--- a/src/cowboy_rest.erl
+++ b/src/cowboy_rest.erl
@@ -130,7 +130,7 @@
 -optional_callbacks([languages_provided/2]).
 
 -callback last_modified(Req, State)
-	-> {calendar:datetime(), Req, State}
+	-> {calendar:datetime() | undefined, Req, State}
 	when Req::cowboy_req:req(), State::any().
 -optional_callbacks([last_modified/2]).
 
-- 

```

essen:
No reason other than not running into use cases where this is useful. Please send a PR. You can mirror what was done for `generate_etag` in defce46fdf85d16bbe3c0a8de6058334e8a53775.

essen:
Merged, thanks!

PR #1689, opened 2025-09-03
Last modified undefined

ThomasArts:
Issue #1686

Make it possible to return `undefined` in REST callback `last_modified` without type error.
This would work with just changing the type, but an extra line of code is added to make it more explicit.

Usecase: With every "store" request one puts a timestamp in some local state in a process. The value to store may go into a database. When this process is restarted, there is an unknown last-modified timestamp.
It might be rather expensive to retrieve this from the database, rather just don't know such that a request to store is needed to get the latest value again.


essen:
Thanks! I expect unrelated errors in CI, something probably broke recently.

ThomasArts:
I tried to run tests locally, but have a lot of needed tools not installed. But it's worrying that so much fails.

PR #1707, opened 2026-02-19
Add HTTP/3 support with pure Erlang QUIC

benoitc:
## Summary

- Add HTTP/3 support using erlang_quic (pure Erlang, no NIF)
- Fix crashes in WebTransport handling
- Add CVE-2019-9514 protection (stream reset rate limiting)
- Implement GOAWAY, CONNECT method, proper error handling

## How to test

Build with QUIC enabled:
```
COWBOY_QUIC=1 make
```

Run all tests:
```
COWBOY_QUIC=1 make eunit
COWBOY_QUIC=1 make ct CT_SUITES=h3
```

Or use the test script:
```
./test_h3.sh
```

### DIFF
--- a/src/cowboy_http3.erl
+++ b/src/cowboy_http3.erl
@@ -19,7 +19,7 @@
 
 -module(cowboy_http3).
 
--ifdef(COWBOY_QUICER).
+-include("cowboy_quic_adapter.hrl").
 
 -export([init/4]).
 
@@ -83,7 +83,7 @@
 -record(state, {
 	parent :: pid(),
 	ref :: ranch:ref(),
-	conn :: cowboy_quicer:quicer_connection_handle(),
+	conn :: ?QUIC_ADAPTER:connection_handle(),
 	opts = #{} :: opts(),
 
 	%% Remote address and port for the connection.
@@ -94,6 +94,9 @@
 
 	%% Client certificate.
 	cert :: undefined | binary(),
+
+	%% HTTP/3 connection status for graceful shutdown.
+	http3_status = connected :: connected | closing,
 
 	%% HTTP/3 state machine.
 	http3_machine :: cow_http3_machine:http3_machine(),
@@ -114,10 +117,14 @@
 
 	%% Streams can spawn zero or more children which are then managed
 	%% by this module if operating as a supervisor.
-	children = cowboy_children:init() :: cowboy_children:children()
+	children = cowboy_children:init() :: cowboy_children:children(),
+
+	%% Stream reset rate limiting (CVE-2019-9514 protection).
+	reset_rate_num = undefined :: undefined | pos_integer(),
+	reset_rate_time = undefined :: undefined | integer()
 }).
 
--spec init(pid(), ranch:ref(), cowboy_quicer:quicer_connection_handle(), opts())
+-spec init(pid(), ranch:ref(), ?QUIC_ADAPTER:connection_handle(), opts())
 	-> no_return().
 
 init(Parent, Ref, Conn, Opts) ->
@@ -126,23 +133,23 @@ init(Parent, Ref, Conn, Opts) ->
 	%% @todo An endpoint MAY avoid creating an encoder stream if it will not be used (for example, if its encoder does not wish to use the dynamic table or if the maximum size of the dynamic table permitted by the peer is zero).
 	%% @todo An endpoint MAY avoid creating a decoder stream if its decoder sets the maximum capacity of the dynamic table to zero.
 	{ok, ControlID} = maybe_socket_error(undefined,
-		cowboy_quicer:start_unidi_stream(Conn, [<<0>>, SettingsBin]),
+		?QUIC_ADAPTER:start_unidi_stream(Conn, [<<0>>, SettingsBin]),
 		'A socket error occurred when opening the control stream.'),
 	{ok, EncoderID} = maybe_socket_error(undefined,
-		cowboy_quicer:start_unidi_stream(Conn, <<2>>),
+		?QUIC_ADAPTER:start_unidi_stream(Conn, <<2>>),
 		'A socket error occurred when opening the encoder stream.'),
 	{ok, DecoderID} = maybe_socket_error(undefined,
-		cowboy_quicer:start_unidi_stream(Conn, <<3>>),
+		?QUIC_ADAPTER:start_unidi_stream(Conn, <<3>>),
 		'A socket error occurred when opening the encoder stream.'),
 	%% Set the control, encoder and decoder streams in the machine.
 	HTTP3Machine = cow_http3_machine:init_unidi_local_streams(
 		ControlID, EncoderID, DecoderID, HTTP3Machine0),
 	%% Get the peername/sockname/cert.
-	{ok, Peer} = maybe_socket_error(undefined, cowboy_quicer:peername(Conn),
+	{ok, Peer} = maybe_socket_error(undefined, ?QUIC_ADAPTER:peername(Conn),
 		'A socket error occurred when retrieving the peer name.'),
-	{ok, Sock} = maybe_socket_error(undefined, cowboy_quicer:sockname(Conn),
+	{ok, Sock} = maybe_socket_error(undefined, ?QUIC_ADAPTER:sockname(Conn),
 		'A socket error occurred when retrieving the sock name.'),
-	CertResult = case cowboy_quicer:peercert(Conn) of
+	CertResult = case ?QUIC_ADAPTER:peercert(Conn) of
 		{error, no_peercert} ->
 			{ok, undefined};
 		Cert0 ->
@@ -150,11 +157,15 @@ init(Parent, Ref, Conn, Opts) ->
 	end,
 	{ok, Cert} = maybe_socket_error(undefined, CertResult,
 		'A socket error occurred when retrieving the client TLS certificate.'),
-	%% Quick! Let's go!
-	loop(#state{parent=Parent, ref=Ref, conn=Conn,
+	%% Initialize rate limiting.
+	CurrentTime = erlang:monotonic_time(millisecond),
+	State0 = #state{parent=Parent, ref=Ref, conn=Conn,
 		opts=Opts, peer=Peer, sock=Sock, cert=Cert,
 		http3_machine=HTTP3Machine, local_control_id=ControlID,
-		local_encoder_id=EncoderID, local_decoder_id=DecoderID}).
+		local_encoder_id=EncoderID, local_decoder_id=DecoderID},
+	State = init_reset_rate_limiting(State0, CurrentTime),
+	%% Quick! Let's go!
+	loop(State).
 
 loop(State0=#state{opts=Opts, children=Children}) ->
 	receive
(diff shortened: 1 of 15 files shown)


leoliu:
Absolutely amazing! This will make HTTP3 much more accessible. Thanks.

benoitc:
I've updated quic to support remaining parts of the spec. All interop tests are passing.

essen:
Hello! Thanks for the PR, I will start looking at it next week. Note however that removing `quicer` is probably too early as we might ship Tanzu RabbitMQ with it this year.

benoitc:
hrm ok i can re-add it. then What do you prefer, having an enviironment variable? ANyway all tests pass now. ANd quic pass all validiation tests. Let me know

essen:
Yes there should be an environment variable. Your early solution of having both `COWBOY_QUICER` and `COWBOY_QUIC` was fine. The default should be no dependency on QUIC for now.

You removed all the HTTP/3, Websocket over HTTP/3 and WebTransport tests as well, they should be added back. You also removed the common `h3` test groups so the PR runs very little HTTP/3 tests right now.

I will enable tests in the PR, just noticed they're waiting approval.

benoitc:
well for me quicer doesn't work at all, their main branch doesn't even compile. Tests are failing. Anyway I will re-add the tests. is RFC 9220 supported by browsers these days ?

essen:
`quicer` is broken against OTP master right now. It works fine with 28 on CI (https://github.com/ninenines/cowboy/actions/runs/22048680378/job/63702146145).

Yes HTTP/3 Websocket is available but only behind flags. It's possible it never gets enabled by default in browsers but browsers aren't the only use case for Websocket.

benoitc:
@essen i've cleaned the patch. I removed the ci changes preventing testing with quicer. This should be OK now.  tests with quic passent.

essen:
Thanks. No action is required at this point on your part, what follows are my thoughts on how to proceed.

So far I have reviewed `erlang_quic` itself. I plan to work on making `erlang_quic` the default, although for now it will be disabled by default. I think the security especially around handshake/auth/verify remains to be proven but this is just a matter of `erlang_quic` becoming more mature / battletested. To be enabled by default it will also need to reach at least `1.0`, and doing this will bump Cowboy to `3.0`.

I expect the performance of `erlang_quic` to not be enough for certain high throughput scenarios. Having to make everything go through a single connection process is definitely a pain point. It is fine for many use cases. But for the few cases where high throughput is important I want to make `quicer` available as a drop-in replacement. My initial goal while reviewing the PR will therefore be how best to keep Cowboy compatible with both of them. This might result in a Ranch-like project specifically for QUIC that would also have Ranch-compatible listener information, where applicable, to make it easier to work with both QUIC and non-QUIC listeners at the same time. The listener start/stop and `cowboy_quicer` types of modules would fit there.

Finally, when `erlang_quic` gets enabled by default, or later, Cowboy probably should have functions to start both TCP and QUIC listeners at the same time with a shared configuration that would automatically set Alt-Svc and friends should the user want that.

I will now review the PR and putting notes for myself (again, no action required for now).

src/cowboy_http3.erl
```diff
@@ -221,13 +239,19 @@ parse(State=#state{opts=Opts}, StreamID, Data, IsFin) ->
...
+					loop(State);
 				false ->
-					%% We avoid logging the data as it could be quite large.
-					cowboy:log(warning, "Received data for unknown stream ~p.",
-						[StreamID], Opts)
-			end,
-			loop(State)
+					%% Stream not found. For erlang_quic, streams are created
+					%% lazily when data arrives (no stream_opened notification).
+					%% Determine stream type from ID and create the stream.
```
essen, line 246: Probably best handled in the adapter (adapter could return two commands/events instead of one).

src/cowboy_http3.erl
```diff
@@ -241,7 +265,13 @@ parse1(State=#state{http3_machine=HTTP3Machine0},
...
 			terminate(State#state{http3_machine=HTTP3Machine}, Error)
 	end;
-%% @todo Handle when IsFin = fin which must terminate the WT session.
+%% Handle FIN on WebTransport session CONNECT stream.
+%% When the CONNECT stream receives FIN, it's equivalent to a clean close.
+parse1(State=#state{conn=Conn}, Stream=#stream{id=SessionID, status=
+		{webtransport_session, _}}, <<>>, fin) ->
+	webtransport_event(State, SessionID, {closed, 0, <<>>}),
+	?QUIC_ADAPTER:shutdown_stream(Conn, SessionID),
+	loop(webtransport_terminate_session(State, Stream));
```
essen, line 274: Changes like this one don't have a test and are unrelated to adding support so will be left for later.

src/cowboy_http3.erl
```diff
@@ -473,11 +503,54 @@ data_frame(State=#state{opts=Opts},
...
 	end.
 
-headers_frame(State, Stream, IsFin, Headers,
-		PseudoHeaders=#{method := <<"CONNECT">>}, _)
+trailers_frame(State, Stream=#stream{status={relaying, _, RelayPid}, id=StreamID}, Trailers) ->
+	RelayPid ! {'$cowboy_relay_trailers', {self(), StreamID}, headers_to_map(Trailers, #{})},
+	stream_store(State, Stream);
+trailers_frame(State=#state{opts=Opts},
+		Stream=#stream{id=StreamID, state=StreamState0}, Trailers) ->
+	try cowboy_stream:info(StreamID, {trailers, headers_to_map(Trailers, #{})}, StreamState0) of
```
essen, line 511: The `trailers` info message is to send response trailers back. Request trailers are currently ignored in Cowboy. Anyway this is not relevant for `erlang_quicer` either.

src/cowboy_http3.erl
```diff
@@ -473,11 +503,54 @@ data_frame(State=#state{opts=Opts},
...
+			Class, Exception, Stacktrace), Opts),
+		reset_stream(State, Stream, {internal_error, {Class, Exception},
+			'Unhandled exception in cowboy_stream:info/3.'})
+	end.
+
+%% Regular CONNECT method (RFC 9110 Section 9.3.6).
+%% Only :method and :authority pseudo-headers are allowed.
+headers_frame(State=#state{ref=Ref, peer=Peer, sock=Sock, cert=Cert},
+		Stream=#stream{id=StreamID}, IsFin, Headers,
+		PseudoHeaders=#{method := <<"CONNECT">>, authority := Authority}, _)
```
essen, line 526: Same, CONNECT is not currently supported and unrelated to the QUIC support, it would need to be added to all protocols at once.

src/cowboy_http3.erl
```diff
@@ -611,7 +684,11 @@ parse_datagram(State, Data0) ->
 			webtransport_event(State, SessionID, {datagram, Data}),
 			loop(State);
 		_ ->
-			error(todo) %% @todo Might be a future WT session or an error.
+			%% Datagram for unknown/terminated session - silently discard.
+			%% This can happen for datagrams arriving for sessions that
+			%% have already been terminated, or for future sessions
+			%% that haven't been established yet.
+			loop(State)
```
essen, line 691: Probably don't want to silence this clause yet.

src/cowboy_http3.erl
```diff
@@ -710,61 +787,27 @@ commands(State0=#state{conn=Conn, http3_machine=HTTP3Machine0},
...
 			State0#state{http3_machine=HTTP3Machine}
 	end,
 	commands(State, Stream, Tail);
-%% Send a push promise.
+%% Server push is not implemented.
+%%
+%% HTTP/2 and HTTP/3 server push has been deprecated and removed by all major
+%% browsers: Chrome removed support in v106 (2022), Firefox in v132 (Oct 2024).
+%% The feature provided minimal real-world benefit and added complexity.
+%% Therefore, push promises are intentionally not supported.
```
essen, line 803: It's still part of the standard and applications may rely on it. Browsers aren't everything. So for now there's no need to decisively drop them (which needs to be done for all protocols at once) nor is there a need to implement them for HTTP/3 (code can stay commented out like it was).

src/cowboy_http3.erl
```diff
@@ -710,61 +787,27 @@ commands(State0=#state{conn=Conn, http3_machine=HTTP3Machine0},
...
-%			State0
-%	end,
-%	commands(State, Stream, Tail);
-%%% Read the request body.
-%commands(State0=#state{flow=Flow, streams=Streams}, Stream, [{flow, Size}|Tail]) ->
+%% Unlike HTTP/2, QUIC handles flow control automatically at the transport layer.
+%% The QUIC stack sends MAX_STREAM_DATA frames to the peer when it's ready to
+%% receive more data. Therefore, the {flow, Size} command is a no-op for HTTP/3.
 commands(State, Stream, [{flow, _Size}|Tail]) ->
-	%% @todo We should tell the QUIC stream to increase its window size.
```
essen: The todo is still valid, we should be able to provide hints to the QUIC stack that we are about to receive larger data sizes. This has improved performance for TCP so there's chances this would for some QUIC implementations as well.

src/cowboy_http3.erl
```diff
@@ -974,49 +1017,85 @@ wt_commands(State0=#state{conn=Conn}, Session=#stream{id=SessionID},
...
-			wt_commands(State, Session, Tail)
-		%% @todo Handle errors.
-	end;
-wt_commands(State0=#state{conn=Conn}, Session=#stream{id=SessionID}, [Cmd|Tail])
+			ok;
+		{error, Reason} ->
+			cowboy:log(warning, "Failed to send WT_DRAIN_SESSION: ~p", [Reason], Opts)
+	end,
+	wt_commands(State, Session, Tail);
+wt_commands(State0=#state{conn=Conn, opts=Opts}, Session=#stream{id=SessionID}, [Cmd|Tail])
```
essen, line 1098: WT changes unrelated AFAICT.

test/rfc9114_quic_SUITE.erl
```diff
@@ -0,0 +1,284 @@
...
+	ok = quic:send_data(Conn, StreamID2, cow_http3:headers(HeaderBlock2), true),
+	_Response1 = do_receive_data(Conn, StreamID1),
+	_Response2 = do_receive_data(Conn, StreamID2),
+	quic:close(Conn, 0),
+	ok.
+
+post_with_body(Config) ->
+	doc("POST request with body that gets echoed back. (RFC9114 4.1)"),
+	Port = config(port, Config),
+	{ok, Conn} = quic:connect("localhost", Port,
```
essen, line 141: As far as Cowboy is concerned it's probably fine to test everything (both `erlang_quic` and `quicer`) with just one or the other. Since `quicer` is better battle tested it's probably a better option initially.

There should be no problem depending on both as well as long as there are no module conflicts.

essen (commented):
No action required on your part.

Supporting both `erlang_quic` and `quicer` in a configurable manner should be easy enough. There are two sides: fetching as dep (no dep should be fetched by default initially) and choosing the backend (we should allow both at the same time in different listeners).

My plan is to create a new Ranch-like dependency that optionally depends on one or two of the QUIC implementations. The initial default in Cowboy will be to not depend on this Ranch-like dependency (similar to what it does with `COWBOY_QUICER`). The dependency will contain what I described in the previous comment.

The PR contains a lot of changes that are not directly related and they are not separate commit so I will keep an eye out on the other changes after adding support for the two QUIC implementations.

essen:
I should have the initial work on the quic/quicer abstraction available some time next week. Putting on the finishing touches now.

src/cowboy_http3.erl
essen: erlang_quic seems to have a `stream_opened` message so perhaps this is dead code?

essen: Not dead code because evidently the message isn't sent, there's only leftovers here and there in erlang_quic.

essen: The lack of a stream_opened event means we can receive a stream_reset or a stop_sending without having received any data first, which is awkward.

PR #1712, opened 2026-03-30
Use Corral for QUIC backend abstraction

essen:
WIP


essen:
With Corral being explicit about the RESET_STREAM/STOP_SENDING difference, Cowboy/Cowlib needs to have its stream close code handling updated for the few remaining tests to pass.

Issue #1672, opened 2025-03-14
Improve linger loop


essen:
https://github.com/erlang/otp/issues/9529#issuecomment-2724280532

We should set `{linger, {true, 0}}` before the loop to ensure that the socket will get closed when the process ends (either at the end of the loop, or when killed).

eagle-head:
Hi @essen, I'd like to take this one.

The fix itself per your OTP comment is mechanical — call `setopts(Socket, [{linger, {true, 0}}])` after `shutdown(write)` in `terminate_linger/1`, mirrored in `cowboy_http2.erl` (same "copied from cowboy_http" block). What I'm unsure about is the test, since the suite is uniformly behavior-driven via CT.

Two shapes I considered:

1. Reproduce the leak from erlang/otp#9529 end-to-end: handler that sends a large body, client reads with a small recv buffer so the server's output buffer stays unacked, kill the connection process, then assert the server-side port is no longer in `erlang:ports()` within a short window. Closest to the actual symptom users hit, but reproduction relies on kernel TCP buffering and may behave differently across Linux/BSD/macOS in CI.

2. Spin a listener with `linger_timeout => infinity`, drive it into the linger loop, and assert via `inet:getopts/2` that the server-side socket has `{linger, {true, 0}}` set. Deterministic but asserts the mechanism rather than the observable behavior.

Is there a precedent in the suite I should follow, or a shape you'd prefer? Also: a single CT case in `http_SUITE` plus the mirrored fix in `cowboy_http2:terminate_linger/1`, or do you want a parallel test in `http2_SUITE` too?

Happy to go in whichever direction you confirm.

essen:
I think we very much want number 1 even if the test doesn't work as well everywhere. If it shows that on Linux without the fix the socket doesn't get closed, and with the fix, it does, then that's good enough. We can leave the test for all platforms we just need it to demonstrate that it improves things.

PR #1715, opened 2026-05-18
Set SO_LINGER to {true, 0} before the linger loop

eagle-head:
Implements the linger fix discussed in #1672 and the related erlang/otp#9529.

When the connection process is killed during the application-level linger loop with outbound data still unacknowledged by the peer, the underlying socket can stay open and the OS port leaks. Setting `SO_LINGER` to `{true, 0}` right after `shutdown(write)` makes the OS send RST and release the connection immediately whenever the BEAM closes the port, instead of waiting on the half-closed state.

Applied in both `cowboy_http:terminate_linger/1` and the mirrored block in `cowboy_http2:terminate_linger/1`.

## Note on testing

An earlier revision included an end-to-end CT case to reproduce the leak, but as @essen pointed out it does not reliably fail without the fix, so it has been removed per his suggestion.

Why it couldn't be made deterministic: the case set `{sndbuf, 2097152}` so the 1 MB body would buffer in the kernel and the process would reach `terminate_linger` with the data still unacknowledged. On Linux, though, an explicit `SO_SNDBUF` is capped by `net.core.wmem_max` (212992 by default) and setting it explicitly also disables send-buffer autotuning, so the real buffer stays well under 1 MB. The body send then blocks inside `cowboy_req:reply`, leaving the process parked in the send rather than in the linger loop — so killing it behaves the same with or without the fix. The fix itself is straightforward, so this PR keeps just the `setopts` change.

Verified locally (Linux, OTP 28):
- `make ct-http` — 35/35 ok
- `make ct-http2` — 18/18 ok

### DIFF
--- a/src/cowboy_http.erl
+++ b/src/cowboy_http.erl
@@ -1603,6 +1603,8 @@ terminate_all_streams(State, [#stream{id=StreamID, state=StreamState}|Tail], Rea
 terminate_linger(State=#state{socket=Socket, transport=Transport, opts=Opts}) ->
 	case Transport:shutdown(Socket, write) of
 		ok ->
+			%% Release the socket on process exit. See erlang/otp#9529.
+			_ = Transport:setopts(Socket, [{linger, {true, 0}}]),
 			case maps:get(linger_timeout, Opts, 1000) of
 				0 ->
 					ok;
--- a/src/cowboy_http2.erl
+++ b/src/cowboy_http2.erl
@@ -1293,6 +1293,8 @@ terminate_all_streams(State, [{StreamID, #stream{state=StreamState}}|Tail], Reas
 terminate_linger(State=#state{socket=Socket, transport=Transport, opts=Opts}) ->
 	case Transport:shutdown(Socket, write) of
 		ok ->
+			%% Release the socket on process exit. See erlang/otp#9529.
+			_ = Transport:setopts(Socket, [{linger, {true, 0}}]),
 			case maps:get(linger_timeout, Opts, 1000) of
 				0 ->
 					ok;


essen:
I am unable to make the test fail without the source changes, waiting on CI to see if it fails there.

essen:
Wait nevermind, CI has the source change so it won't fail there.

Can you tell me how you ran the test to confirm that it failed without the source code changes?

If the test doesn't fail before the source code change then we should just not have it. It's OK if the test is too difficult to create and the change is obvious.

eagle-head:
You're right — I can't give you a dependable pre-fix failure recipe, so I've removed the test and kept just the `setopts` fix.

What was happening: the case set `{sndbuf, 2097152}` expecting the 1 MB body to buffer in the kernel so the process would reach `terminate_linger` with the data still unacked. But on Linux an explicit `SO_SNDBUF` is capped at `net.core.wmem_max` (212992 here → `inet:getopts` reports 425984) and setting it explicitly also disables send-buffer autotuning, so the real buffer is well under 1 MB. The body send blocks inside `cowboy_req:reply`, so the process is parked in the send — not in `terminate_linger` — when the test kills it; the `{linger, {true, 0}}` line never runs, so the kill behaves the same with or without the fix. The red→green I reported earlier was environment/timing dependent, not the mechanism the test describes.

So I've dropped `terminate_linger_releases_socket_on_kill` and the `long_body_h` handler; the PR is now just the two-line fix in `cowboy_http`/`cowboy_http2`. Happy to add the deterministic `inet:getopts` check instead if you'd prefer some coverage.

PR #1721, opened 2026-08-20
HTTP/1.1: Reject non-token characters in header names

lalalala5678:
RFC 9110 defines `field-name = token`. Previously Cowboy only rejected whitespace, CR, LF and NUL bytes in header field names and accepted all other bytes (control characters, separators, high bytes) via the catch-all clause in `parse_hd_name/4`.

This adds the `when ?IS_TOKEN(C)` guard on the accepting clause, as you suggested by email following my report. Any byte that is not a valid token character is now rejected with a 400 status code and the connection is closed, matching the existing parse error handling (same mechanism as the NUL byte clause). The specific error clauses for whitespace, CR/LF and NUL still fire with their dedicated reasons.

A test is added to `rfc7230_SUITE` covering control characters, separators, high bytes, an invalid first character and an invalid character immediately before the colon. Full `make ct` run: rfc7230_SUITE 158/158 ok; no regressions elsewhere (the only failures are `static_handler_SUITE:dir_error_unreadable`, which fails when the test suite runs as root, and intermittent timing-sensitive `sys_SUITE` cases — both reproduce on unmodified master).


lalalala5678:
CI note: the failing jobs are unrelated to this change. They are (1) a Windows pacman mirror failure downloading ca-certificates (tests never started), and (2) timing-sensitive flakes in `proxy_header_SUITE`/`req_SUITE`. The same job matrix has been failing on `master` scheduled runs since 2026-08-03 (see runs on Aug 3/10/17, identical failing set). `rfc7230_SUITE` — the suite covering this change — passes on all platforms, and Dialyzer is clean.

